PR #1, opened 2016-02-06
update default url param and model

leb2dg:
# Changes
- update the default query url for the experiment page so that it is not using elastic search
- add necessary fields to experiment model
  - active (possibly change to status?) can be 'Active', 'Draft', or 'Archive'
# Question
- is date the proper type for `endDate`, `beginDate`, and `lastEdited`

### DIFF
--- a/exp-models/addon/models/experiment.js
+++ b/exp-models/addon/models/experiment.js
@@ -7,10 +7,13 @@ import JamModel from '../mixins/jam-model';
 
 export default DS.Model.extend(JamModel, {
     title: DS.attr('string'),
-    description: DS.attr('string'),
-    active: DS.attr('boolean'),
-    structure: DS.attr(),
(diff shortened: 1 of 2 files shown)


exp-models/addon/models/experiment.js
```diff
@@ -7,10 +7,13 @@ import JamModel from '../mixins/jam-model';
 
 export default DS.Model.extend(JamModel, {
     title: DS.attr('string'),
-    description: DS.attr('string'),
-    active: DS.attr('boolean'),
-    structure: DS.attr(),
-    permissions: DS.attr(),
+	description: DS.attr('string'),
```
abought: Very very tiny, but could you indent 4 spaces for consistency with our code style conventions elsewhere?

samchrisinger: +1

PR #3, opened 2016-02-08
Fix issues with session record

abought:
Intended as companion to https://github.com/CenterForOpenScience/experimenter/pull/17

Refs https://openscience.atlassian.net/browse/LEI-42

## Purpose
Various changes to session model, needed to support related PR.

## Summary of changes
- DS.store should be DS.attr (fixes issue causing ember inspector to crash)
- Field name and relationship changes
- Apply query fix to all document records
- Add ability to track a text "<relationAttr>Id" field to a record, for any constructed records. Should save an API call when we want title of a related record.

### DIFF
--- a/exp-models/addon/mixins/jam-document-adapter.js
+++ b/exp-models/addon/mixins/jam-document-adapter.js
@@ -3,7 +3,7 @@ import Ember from 'ember';
 export default Ember.Mixin.create({
     findRecordUrlTemplate: '{+host}/{+namespace}/documents{/id}',
     findAllUrlTemplate: '{+host}/{+namespace}/collections{/jamNamespace}.{+collectionId}/documents',
-    queryUrlTemplate: '{+host}/{+namespace}/collections{/jamNamespace}.{+collectionId}/_search',
+    queryUrlTemplate: '{+host}/{+namespace}/collections{/jamNamespace}.{+collectionId}/documents',
 
     // TODO: Support creation and deletion of records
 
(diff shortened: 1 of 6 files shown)


exp-models/addon/mixins/jam-document-adapter.js
```diff
@@ -3,7 +3,7 @@ import Ember from 'ember';
 export default Ember.Mixin.create({
     findRecordUrlTemplate: '{+host}/{+namespace}/documents{/id}',
     findAllUrlTemplate: '{+host}/{+namespace}/collections{/jamNamespace}.{+collectionId}/documents',
-    queryUrlTemplate: '{+host}/{+namespace}/collections{/jamNamespace}.{+collectionId}/_search',
+    queryUrlTemplate: '{+host}/{+namespace}/collections{/jamNamespace}.{+collectionId}/documents',
```
samchrisinger: Can you explain the need for this change? Lauren expects to need to use ES's full-text search for her work.

abought: The same base query URL should be used across all documents (Lauren's original fix was for only experiments).

Ostriches proposed an elegant compromise which takes advantage of the similar parts of the URL, and will allow us to do both searching and querying by making this URL dynamic. I think @leb2dg  is working on that change in a followup PR.

samchrisinger: Thanks for the explanation.

PR #8, opened 2016-02-15
Bcrypt new passwords before sending them to the server

chrisseto:


exp-models/addon/serializers/account.js
```diff
@@ -3,6 +3,21 @@ import DS from 'ember-data';
 import JamSerializer from '../mixins/jam-serializer';
 import JamDocumentSerializer from '../mixins/jam-document-serializer';
 
+let bcrypt = dcodeIO.bcrypt;
+
 export default DS.JSONAPISerializer.extend(JamSerializer, JamDocumentSerializer, {
     modelName: 'account',
+    relationAttrs: ['sessions'],
```
samchrisinger: Did you add this back on purpose? I think @abought removed it.

chrisseto: Merge conflict, sorry about that.

PR #9, opened 2016-02-17
 [WIP] (models) Use different sessions collection per experiment

abought:
Refs https://openscience.atlassian.net/browse/LEI-78

## Purpose
See detailed notes on companion PR, https://github.com/CenterForOpenScience/experimenter/pull/31

### DIFF
--- a/exp-models/addon/mixins/jam-document-adapter.js
+++ b/exp-models/addon/mixins/jam-document-adapter.js
@@ -16,8 +16,7 @@ export default Ember.Mixin.create({
         },
         collectionId: (type /*, id, snapshot, query*/) => Ember.Inflector.inflector.pluralize(type)
     },
-    ajax: function(url, type, options) {
-        options = options || {};
+    ajax: function(url, type, options={}) {
         options.traditional = true;
         return this._super(...arguments);
     }
--- a/exp-models/addon/mixins/jam-document-adapter.js
+++ b/exp-models/addon/mixins/jam-document-adapter.js
@@ -17,6 +17,7 @@ export default Ember.Mixin.create({
         collectionId: (type /*, id, snapshot, query*/) => Ember.Inflector.inflector.pluralize(type)
     },
     ajax: function(url, type, options) {
+        options = options || {};
         options.traditional = true;
         return this._super(...arguments);
     }
--- a/exp-models/addon/mixins/jam-document-serializer.js
+++ b/exp-models/addon/mixins/jam-document-serializer.js
@@ -10,7 +10,7 @@ export default Ember.Mixin.create({
         modifiedBy: {serialize: false},
     },
 
-    payloadKeyFromModelName: function(modelName) {
+    payloadKeyFromModelName: function(/*modelName */) {
         // JamDB expects all collections to specify JSONAPI type 'documents'
         return 'documents';
     },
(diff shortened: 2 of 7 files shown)


exp-models/addon/mixins/jam-document-adapter.js
```diff
@@ -17,6 +17,7 @@ export default Ember.Mixin.create({
         collectionId: (type /*, id, snapshot, query*/) => Ember.Inflector.inflector.pluralize(type)
     },
     ajax: function(url, type, options) {
```
samchrisinger: Just fyi es6 supports: `ajax: function(url, type, options={}) {`

exp-models/addon/models/experiment.js
```diff
@@ -21,14 +26,31 @@ export default DS.Model.extend(JamModel, {
...
+
+    _registerSessionModels() {
+        // Dynamically register the required models for a session table associated with this experiment
+        var cId = this.get('sessionCollectionId');
+        window.App.register(`model:${cId}`, SessionModel.extend()); // register a dummy model. This seems to work even if model already registered
+        window.App.register(`adapter:${cId}`, SessionAdapter.extend({sessionCollectionId: cId})); // Override part of adapter URL
+        window.App.register(`serializer:${cId}`, SessionSerializer.extend({modelName: cId})); // Tell serializer what model to use)
+    },
+
+    init() {
```
samchrisinger: Do you need a super call here?

exp-models/addon/mixins/jam-document-adapter.js
abought: O' frabjous day! Callooh! Callay!
(changed this and verified working)

exp-models/addon/models/experiment.js
abought: There wasn't a parent method to call- but there might be. Good catch.

exp-models/addon/mixins/jam-document-serializer.js
```diff
@@ -10,7 +10,7 @@ export default Ember.Mixin.create({
         modifiedBy: {serialize: false},
     },
 
-    payloadKeyFromModelName: function(modelName) {
+    payloadKeyFromModelName: function(/*modelName */) {
         // JamDB expects all collections to specify JSONAPI type 'documents'
         return 'documents';
```
chrisseto: What about collections?

samchrisinger: see `jam-document-serializer`

exp-models/addon/models/experiment.js
```diff
@@ -21,14 +26,32 @@ export default DS.Model.extend(JamModel, {
...
+    sessionCollectionId: Ember.computed('shortId', function() {
+        // Return a string corresponding to the session collection shortID, to be used by model/adapter/serializer
+        // Eg an experiment called 'test0' would have a collection 'session-test0'
+        return `session${this.get('shortId')}s`; // FIXME: the spurious s is a requirement imposed by genschemas...
+    }),
+
+    _registerSessionModels() {
+        // Dynamically register the required models for a session table associated with this experiment
+        var cId = this.get('sessionCollectionId');
+        window.App.register(`model:${cId}`, SessionModel.extend()); // register a dummy model. This seems to work even if model already registered
```
chrisseto: Is there a better workaround to this?  Using `window` in a built app makes me uneasy

abought: It makes me a little worried too. 

It seems that Ember [removed](http://emberigniter.com/accessing-global-object-in-ember-cli-app/) the global app reference it used to have, and this approach seems to depend on calling `application.register`... I honestly don't know if there's another way to get at the app instance.

This is about all I could find on the lifetime of the window object, but that's a thin shield. I agree I'd like to see a better way.

PR #14, opened 2016-02-22
Add eligibility helpers

samchrisinger:
Changes:

- Add eligibility.js utils for parsing and checking eligibility
- Make account.profiles computed to allow transformation into a genuine Ember.Object subclass
- Add 'age' computed to Profile class

### DIFF
--- a/exp-models/addon/models/account.js
+++ b/exp-models/addon/models/account.js
@@ -2,6 +2,9 @@
 Manage data about one or more documents in the accounts collection
  */
 
+import moment from 'moment';
+
+import Ember from 'ember';
 import DS from 'ember-data';
 import JamModel from '../mixins/jam-model';
 
@@ -20,6 +23,15 @@ export default DS.Model.extend(JamModel, {
         var getProfile = function(item) {
             return item.profileId === profileId;
         };
-        return profiles.filter(getProfile)[0];
+
+        var Profile = Ember.Object.extend({
+            age: Ember.computed('birthday', function() {
+                var bd = moment(this.get('birthday'));
+                return moment(new Date()).diff(bd, 'days');
+            })
+        });
+        return Profile.create(
+            profiles.filter(getProfile)[0]
+        );
     }
 });
--- a/exp-models/addon/models/account.js
+++ b/exp-models/addon/models/account.js
@@ -9,6 +9,7 @@ import DS from 'ember-data';
 import JamModel from '../mixins/jam-model';
 
 var Profile = Ember.Object.extend({
+    birthday: null,
     age: Ember.computed('birthday', function() {
         var bd = moment(this.get('birthday'));
         return moment(new Date()).diff(bd, 'days');
@@ -26,7 +27,6 @@ export default DS.Model.extend(JamModel, {
             });
         },
         set: function(value) {
-            debugger;
             this.set('_profiles', value);
         }
     }),
@@ -38,11 +38,9 @@ export default DS.Model.extend(JamModel, {
         // Scan the list of profiles and gets first one with matching ID (else undefined). Assumes profileIds are unique.
         var profiles = this.get('profiles') || [];
         var getProfile = function(item) {
-            return item.profileId === profileId;
+            return item.get('profileId') === profileId;
         };
 
-        return Profile.create(
-            profiles.filter(getProfile)[0]
-        );
+        return profiles.filter(getProfile)[0];
     }
 });
--- a/exp-models/addon/models/account.js
+++ b/exp-models/addon/models/account.js
@@ -8,11 +8,28 @@ import Ember from 'ember';
 import DS from 'ember-data';
 import JamModel from '../mixins/jam-model';
 
+var Profile = Ember.Object.extend({
+    age: Ember.computed('birthday', function() {
+        var bd = moment(this.get('birthday'));
+        return moment(new Date()).diff(bd, 'days');
+    })
+});
+
 export default DS.Model.extend(JamModel, {
     username: DS.attr('string'),
     password: DS.attr('string'),
-    profiles: DS.attr(),  // TODO: Pages using profiles will be fragile unless there is a clear, explicit contract of what fields ember can always expect to be present. This nested document doesn't inherently specify as such.
-
+    _profiles: DS.attr(), // TODO: Pages using profiles will be fragile unless there is a clear, explicit contract of what fields ember can always expect to be present. This nested document doesn't inherently specify as such.
+    profiles: Ember.computed('_profiles', {
+        get: function() {
+            return (this.get('_profiles') || []).map((p) => {
+                return Profile.create(p);
+            });
+        },
+        set: function(value) {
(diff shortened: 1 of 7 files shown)


exp-models/addon/models/account.js
```diff
@@ -2,14 +2,34 @@
 Manage data about one or more documents in the accounts collection
  */
 
+import moment from 'moment';
```
abought: Is this line necessary, given that we use `app.import` in `ember-cli-build.js`?

https://guides.emberjs.com/v2.0.0/addons-and-dependencies/managing-dependencies/#toc_globals-provided-by-javascript-assets

exp-models/addon/models/account.js
```diff
@@ -2,14 +2,34 @@
...
+        return moment(new Date()).diff(bd, 'days');
+    })
+});
+
 export default DS.Model.extend(JamModel, {
     username: DS.attr('string'),
     password: DS.attr('string'),
-    profiles: DS.attr(),  // TODO: Pages using profiles will be fragile unless there is a clear, explicit contract of what fields ember can always expect to be present. This nested document doesn't inherently specify as such.
-
+    _profiles: DS.attr(), // TODO: Pages using profiles will be fragile unless there is a clear, explicit contract of what fields ember can always expect to be present. This nested document doesn't inherently specify as such.
```
abought: I understand the desire to have an ember object here, but aliasing the field can get very... messy.

It would be cleaner to name the computed observable field something completely different, like `profileData`. That would let us eliminate the custom serializer logic, and avoid any confusion when someone tries to use `createRecord` based on the layout of the JSON document.

(When individual fields break the rules on a case-by-case basis, it can get very confusing, especially since otherwise our models are explictly designed to look like the server fields wherever possible)

exp-models/addon/models/account.js
```diff
@@ -2,14 +2,34 @@
...
+    profiles: Ember.computed('_profiles', {
```
abought: Come to think of it- is there any reason we can't just use a custom `DS.Transform` in place of the computed field?

You'd want to test it using `createRecord` (to ensure that something that is already a plain object  would serialize correctly). The idea would be to define the field as: `profiles = DS.attr('customprofiletransformpowersactivate'`)
http://emberjs.com/api/data/classes/DS.Transform.html

exp-models/addon/models/experiment.js
```diff
@@ -19,9 +20,9 @@ export default DS.Model.extend(JamModel, {
 
     title: DS.attr('string'),
     description: DS.attr('string'),
-    beginDate: DS.attr('date'),	// TODO: ISODate
-    endDate: DS.attr('date'),	// TODO: ISODate
-    lastEdited: DS.attr('date'),	// TODO: ISODate
+    beginDate: DS.attr('date'), // TODO: ISODate
```
abought: These TODO lines are unnecessary- modern versions of ember use ISO date format already without the need for further transforms. 

(I could have sworn I removed those comments before, but I guess not?)

exp-models/addon/serializers/account.js
```diff
@@ -7,15 +7,22 @@ let bcrypt = dcodeIO.bcrypt;
...
-
     serialize(record, options) {
-      if(record.record.get('isNew')) {
-        record = record.record;
-        var salt = bcrypt.genSaltSync(12);
-        record.set('password', bcrypt.hashSync(record.get('password'), salt));
-        record = record._createSnapshot();
-      }
+        record.profiles = record._profiles;
+        delete record._profiles;
```
abought: See note above about renaming the computed. We may be able to simplify the serializer logic.

exp-models/addon/utils/eligibility.js
```diff
@@ -0,0 +1,113 @@
...
+    return function() {
+        return a(...arguments) || b(...arguments);
+    };
+};
+
+/**
+ * @return {integer}: number of days represented by amount/unit pair, e.g. 3 'weeks' -> 21
+ **/
+function parseDate(amount, unit='days') {
+    var singular = unit.replace(/s$/, '');
```
abought: I would sugggest using `Ember.Inflector.inflector.singularize(word)`, or just skipping this line and perhaps using `pluralize` directly. (you seem to be making something singular in order to pluralize)
http://emberjs.com/api/data/classes/Ember.Inflector.html

Ember's inflector is designed to handle more examples of irregular plurals, and using it would let us reuse any words that had already been registered as `uncountable` etc. 

Basically, just a matter of standardizing on existing logic instead of rewriting our own. :)

exp-models/addon/utils/eligibility.js
```diff
@@ -0,0 +1,113 @@
...
+    return moment.duration(parseFloat(amount), `${singular}s`).asDays();
+}
+
+function compileEligibilityString(elig) {
+    var parts = elig.split(/(\(|\)|and|or)/).map((p) => p.trim()).filter((p) => p.length);
```
abought: If I am understanding this right: it appears you have defined a grammar or mini-DSL for specifying eligibility criteria. Can you point me to an example file containing a usage, or minimum documentation on the intended clauses? There does not appear to be any new experiment test data in https://github.com/CenterForOpenScience/experimenter/pull/38

I'd like to capture basic design notes as we merge PRs, rather than try to reconstruct from memory later. A 2-3 sentence description + example would be a valuable addition to the PR description.

(Sorry for the extra work. You can call me anything on this list in return: https://mraybould.wordpress.com/2013/01/11/shakespeare-insult-kit/ )

samchrisinger: https://github.com/CenterForOpenScience/experimenter/blob/develop/scripts/data/experiment.json#L9

But yes, I will add docs here

exp-models/addon/utils/eligibility.js
```diff
@@ -0,0 +1,129 @@
...
+/**
+ * @return {integer}: number of days represented by amount/unit pair, e.g. 3 'weeks' -> 21
+ **/
+function parseDate(amount, unit='days') {
+    var singular = unit.replace(/s$/, '');
+    return moment.duration(parseFloat(amount), `${singular}s`).asDays();
+}
+
+/** Compile an eligibiliy string into a function: (participant) -> boolean
+ *
```
abought: Ah, example was added before this PR. Thanks!

Since I don't see an easy way to validate this with a simple regex, we'll definitely want to keep this documentation when it comes time for users to define experiments.

exp-models/addon/models/account.js
samchrisinger: Feels creepy not to add the import statement here, and the ember-moment docs show this as an example: https://github.com/stefanpenner/ember-moment#es6-moment

So probably no.

abought:
Pass complete. There are a few places we can simplify the lgoic, which I'd recommend before merge if possible.

Although I'm a bit leery about the custom DSL, it's probably what we need to move forward for now. Thanks for the documentation! (including where the custom units are drawn from; very helpful)

PR #18, opened 2016-02-24
Fix state and component transition issues

abought:
Refs https://openscience.atlassian.net/browse/LEI-88
Tracking PR; see checklist for status.

## Purpose
Fix issues of shared and lingering state in the experimenter player.

- [x] Same timings data is replicated across all frames

- [x] Clicking the last consent frame: problem where button does not correctly activate until checked and unchecked several times

- [ ] When two video player frames are shown in a row, the second frame does not load the new video. It simply uses the video content from the first video frame. (even though the poster image loads correctly, it does not even attempt to download the video piece)
  - It appears that `willDestroyElement` is not called either. Speculatively, it seems that the *same* component is getting new parameters; rather than creating a second new instance of the same type.

## Summary of changes
- Make sure that init creates instance variables for  `meta.data` (things to be serialized), in addition to the vars already added for `meta.parameters` (things to control config of page)

### DIFF
--- a/exp-player/addon/components/exp-audioplayer.js
+++ b/exp-player/addon/components/exp-audioplayer.js
@@ -43,18 +43,17 @@ export default ExpFrameBaseComponent.extend({
                     default: []
                 }
             }
-        }
-    },
-
-    data: {
-        type: 'object',
-        properties: {  // We don't *need* to tell the server about this, but it might be nice to track usage of the setup page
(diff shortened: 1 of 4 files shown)


abought:
Advancing this PR to candidate status with 2 of the 3 fixes; will have to address the double-component bug separately at a later date.

PR #23, opened 2016-02-28
Add experiment thumbnails

samchrisinger:


exp-models/addon/models/experiment.js
```diff
@@ -74,5 +83,27 @@ export default DS.Model.extend(JamModel, {
...
         collection.save();
-    }.on('didCreate')
+    }.on('didCreate'),
+
+    setThumbnail(raw) {
+        var self = this;
+
+        var thumbnailId = this.get('thumbnailId');
+        if (thumbnailId) {
+            return this.get('store').findRecord('thumbnail', thumbnailId)
```
abought: Can you reuse the `thumbnail` computed property to DRY this code?
https://github.com/CenterForOpenScience/exp-addons/pull/23/files#diff-2d6e780c9343997d7d3e8bcbe18f507fR28

exp-models/addon/models/experiment.js
```diff
@@ -24,6 +24,15 @@ export default DS.Model.extend(JamModel, {
     endDate: DS.attr('date'),
     structure: DS.attr(),
 
+    thumbnailId: DS.attr('string'),
```
abought: Design question: this field seems to impose a 1:1 relationship between experiments and thumbnails.

Given that, can you help clarify the reason we choose to store thumbnails in a separate collection, rather than as part of the experiment?

abought: Discussed in person with Sam. It seems the separate collection was a requirement to prevent elasticsearch from indexing this data. 

It might help to add a brief source code comment to this effect (to keep future maintainers from overzealous refactorings), but not essential.

exp-models/addon/models/experiment.js
```diff
@@ -29,6 +29,40 @@ export default DS.Model.extend(JamModel, {
...
+            if (thumbnailId) {
+                return this.get('store').findRecord('thumbnail', thumbnailId);
+            }
+            return null;
+        },
+        set (raw) {
+            var self = this;
+
+            var thumbnailId = this.get('thumbnailId');
+            if (thumbnailId) {
```
abought: Using computed getters and setters is a nice idea. :) Just to go further- can the setter call the getter value?

(specifically, I'm noticing that the store.findRecord code is repeated)

Not a huge issue if it works.

PR #19, opened 2016-02-25
Prototype random frame (player) 

abought:
Refs https://openscience.atlassian.net/browse/LEI-89
Companion to https://github.com/CenterForOpenScience/experimenter/pull/41

See companion PR for full description.

PR #25, opened 2016-03-01
Feature/video

samchrisinger:


exp-player/addon/components/exp-player.js
```diff
@@ -4,12 +4,13 @@ import parseExperiment from '../utils/parse-experiment';
 
 export default Ember.Component.extend({
     layout: layout,
+    store: Ember.inject.service('store'),
```
abought: This was removed as part of the experiment refactor:
https://github.com/CenterForOpenScience/exp-addons/pull/19/files#diff-85b6be66a5ce62f39ee05c9d04ceb67fL6

abought: See comments below: as ember stands at present, it's not a great idea to make components own data or interact with the store directly. (hence the refactor moving that responsibility into the route, and saving via an action) 

Once I've had a chance to review the rest of the PR we can discuss concrete ways to work the new code into that scheme.

exp-player/addon/components/exp-video-config.js
```diff
@@ -0,0 +1,46 @@
+import ExpFrameBaseComponent from 'exp-player/components/exp-frame-base';
+import layout from 'exp-player/templates/components/exp-video-config';
+
+export default ExpFrameBaseComponent.extend({
+    layout,
+    videoRecorder: Em.inject.service(),
```
abought: Do you need to import Ember in this file for this to work?

abought:
Note that the sample data in the companion PR will need some (very small) reorganization to work with the new experiment schema.
https://github.com/CenterForOpenScience/experimenter/pull/46

Old: `experiment.structure = [frameConfigs]`
New:
```
experiment.structure = {
  frames: 
    {'firstId': firstFrameConfig} , 
  sequence: ['firstId']
}
```

exp-player/addon/components/exp-video-consent.js
```diff
@@ -0,0 +1,61 @@
...
+            }
+        },
+        data: {
+            type: 'object',
+            properties: {
+                consentGranted: {
+                    type: 'boolean',
+                    default: false
+                },
+                videoId: {
```
abought: How does this information make it from the video recorder, to the component controller? I'm probably just missing something here.

(the items in `data` are what we expect to be sent to the server, though in the absence of validation this is more a notation convention than a rule)

exp-player/index.js
```diff
@@ -1,13 +1,339 @@
...
+  //desc: This controls whether or not HDFVR will trigger server side the execution of FFMPEG converting once the stream finished uploading.
+  //values:1 for enabled, 0 for disasbled
+  //default: 0 (disabled)
+  enableFFMPEGConverting: 0,
+
+  //ffmpegCommand:String
+  //desc: This sets the conversion command that will be executed with FFMPEG when enableFFMPEGConverting is set to 1. The command has the following pattern: "ffmpeg -i [THE_INPUT_FLV_FILE]  [codec parameters audio/video] [THE_OUTPUT_MP4_FILE]" . The [THE_INPUT_FLV_FILE] and [THE_OUTPUT_MP4_FILE] parts must not be changed, both the path to ffmpeg and to the video files will be detected and set automatically by the media server. Only the codec parameters will be taken into account. The command is expressed like this to put it more into context as opposed to just sending the codec parameters.
+  //values:Example command: ffmpeg -i [THE_INPUT_FLV_FILE] -c:v libx264 [THE_OUTPUT_MP4_FILE]
+  //default: ffmpeg -i [THE_INPUT_FLV_FILE] -c:v libx264 [THE_OUTPUT_MP4_FILE]
+  ffmpegCommand: "ffmpeg -i [THE_INPUT_FLV_FILE] -c:v libx264 [THE_OUTPUT_MP4_FILE]",
```
abought: Does this code need to be in the front-end?

abought:
Review-for-regressions pass complete; other than what #26 fixed I think we got most of the thorny bits. Thanks Sam!

 The biggest change remaining is the need to separate data handling logic out of the component. (see `controllers/testplayer.js` and `routes/testplayer.js` for code to generate a blank session- this PR was using some very old code for saving data, which did not let us check for unsaved data on `willTransition`)

Once that and the sample data in https://github.com/CenterForOpenScience/experimenter/pull/46 are up to date, I'd be happy to serve as a guinea pig for functional testing. Let me know how I can help.

exp-player/addon/components/exp-player.js
```diff
@@ -18,7 +19,19 @@ export default Ember.Component.extend({
         var frameConfigs = parseExperiment(this.get('experiment.structure'));
         this.set('frames', frameConfigs);  // When player loads, convert structure to list of frames
     },
-
+    session: Ember.computed('experiment', function() {
+        var session = this.get('store').createRecord(this.get('experiment.sessionCollectionId'), {
```
samchrisinger: My understanding of the Ember docs is that by default computed values are cached: http://emberjs.com/api/classes/Ember.ComputedProperty.html#method_volatile.

So the use of a computed here is very deliberate. If the experiment changes, the session should change.

exp-player/index.js
chrisseto: Yep

exp-player/addon/components/exp-video-config.js
chrisseto: Yes it should. Looks like `Em` and `Ember` are globals so it works anyways

abought: Ah, so the jshint error can be suppressed then.

`exp-player/addon/components/exp-video-config.js: line 6, col 20, 'Em' is not defined.`

(see "ignore" on this page: http://jshint.com/docs/ )

chrisseto: I'd prefer just adding importing `Em`. Up to @samchrisinger

PR #30, opened 2016-03-02
Add pastSessions context

samchrisinger:
# Purpose

Some frames in the exp-player will need to know about past times a user has taken a given experiment. This creates a `currentFrameContext` in the exp-player and passes that context into each frame. This context includes a list of all past sessions.

Additionally this updates exp-exit-survey as a demonstration of this approach.

This also adds a getCurrentUser service that currently uses hard coded values to fetch one of our sample acconts/profiles.


exp-models/addon/mixins/jam-document-adapter.js
```diff
@@ -3,7 +3,9 @@ import Ember from 'ember';
 export default Ember.Mixin.create({
     findRecordUrlTemplate: '{+host}/{+namespace}/documents{/id}',
     findAllUrlTemplate: '{+host}/{+namespace}/collections{/jamNamespace}.{+collectionId}/documents',
+
     queryUrlTemplate: '{+host}/{+namespace}/collections{/jamNamespace}.{+collectionId}{/search}',
+    queryRecordUrlTemplate: '{+host}/{+namespace}/collections{/jamNamespace}.{+collectionId}{/search}',
```
abought: Thanks- I keep forgetting to add this in.

exp-player/addon/components/exp-exit-survey.js
```diff
@@ -123,37 +138,47 @@ export default ExpFrameBaseComponent.extend({
...
+        return (this.get('frameContext.pastSessions') || []).length;
+    }),
+    currentDaysSessionsCompleted: Ember.computed('frameContext', function() {
+        var pastSessionDates = this.get('frameContext.pastSessions').map((session) => {
+            return moment(session.get('createdOn'));
+        });
+        var minDate = moment.min(pastSessionDates);
+        var maxDate = moment.max(pastSessionDates);
+
+        return maxDate.diff(minDate, 'days');
```
abought: How do we define days completed?
a. Number of days between two specific sessions, or
b. The number of days that the participant was actually present for

This seems to implement (a), but option (b) feels more like what the experimenter would want to track.

brianjgeiger: This is for knowing when to let someone do another session of their longitudinal? I would presume A (they would need, for example, a week between sessions). If something else please disregard.

abought: From discussion with Sam: this will require action (a followup question with the client, perhaps). A TODO or even a JIRA ticket would be recommended to make sure that we don't forget to come back and fix this feature later.

Also, based on which definition we choose, it might help to rename this (or add a comment) to make the purpose and definition more clear.

abought:
I was unable to run the code on this paired branch until I ran `bower install swfobject` in the top-level `experimenter` folder.

exp-player/addon/components/exp-exit-survey.js
```diff
@@ -95,6 +97,20 @@ export default ExpFrameBaseComponent.extend({
                     type: 'jsonschema',
                     description: 'JSON-schema defining second form.',
                     default: emailOptOutSchema
+                },
+                idealSessionsCompleted: {
```
abought: From a reusability standpoint, having these default values live in the component would seem to lock us to the assumption that every use of this component is a longitudinal study.

It might be better to use default values that trigger no behavior at all, rather than hardcoding specific default numbers.

exp-player/addon/services/get-current-user.js
```diff
@@ -0,0 +1,11 @@
+import Ember from 'ember';
+
+export default Ember.Service.extend({
+    store: Ember.inject.service(),
+    load: function() {
+        // TODO work for real
+        return this.get('store'). findRecord('account', 'experimenter.accounts.tester0').then((account) => {
```
abought: The account model provides a helper function called `profileById` which may be useful for implementing this for real. https://github.com/abought/exp-addons/blob/ce9c7a094082ac82f9d35932eca71adb99231167/exp-models/addon/models/account.js#L18-L18

Separately, if we are not planning to implement this immediately (based on reviewing this for merge before it is there), a JIRA ticket might help to remind us and track this needed element of work.

exp-player/addon/components/exp-exit-survey.js
```diff
@@ -123,37 +138,47 @@ export default ExpFrameBaseComponent.extend({
...
                 var formData = root.get('formData');
                 Ember.merge(formData, this.getValue());
                 root.set('formData', formData);
-                console.log('Post-study survey complete.')
                 root.actions.next();
             }
-        }
+        };
+    }),
+    currentSessionsCompleted: Ember.computed('frameContext', function() {
```
abought: Do we think that other frames would also need (optional) access to session information? If so, perhaps this code could be moved to the base frame or a mixin?

exp-player/addon/components/exp-player.js
```diff
@@ -11,6 +11,21 @@ export default Ember.Component.extend(FullScreen, {
...
+    store: Ember.inject.service(),
+    getCurrentUser: Ember.inject.service(),
+    pastSessions: Ember.computed('experiment', function() {
+        return this.get('getCurrentUser').load().then(([, profile]) => {
+            var profileId = profile.get('profileId');
+            var experimentId = this.get('experiment.id');
+            return this.get('store').queryRecord(this.get('experiment.sessionCollectionId'), {
+                q: [
+                    `experimentId:${experimentId}`,
+                    `profileId:${profileId}`
```
abought: It's worth noting that this returns a promise, not a value. Usually that should be ok, but (especially since this is tied to other properties that recompute when it changes), it's possible you may want to wait until the promise is resolved to return a value.

abought: Also, what is the rationale for putting store queries (for pastSessions) on the player, instead of on the route? It seems weird to have responsibility for prequisite data live in more than one place.

abought:
Possible bug during functional review. After running the bootstrap script and updating all dependencies, I ran through the following recommended experiment page in chrome and clicked "submit" when done with the survey:

Experiment: http://localhost:4200/preview/experimenter.experiments.test1

The result was a blank page and a console error `Uncaught TypeError: this.get(...).map is not a function` 
(exp-exit-survey.js:169)

abought:
Paused review for tonight. There are two features whose implementation may be incomplete, plus a few questions about where the data and code should live (organization etc). I'll try to touch base in the AM about the console errors I was seeing.

exp-player/addon/components/exp-exit-survey.js
samchrisinger: This frame fulfills a very specific use case. We are ignoring the general case for now.

exp-player/addon/components/exp-exit-survey.js
samchrisinger: This is a computed value of the frameContext (available to every frame).  

If we find a more general use case we can move this code to a mixin. For now it's fine here.

samchrisinger:
The mentioned bug is an existing issue with the exp-exit-frame implementation. Resolved in: https://github.com/CenterForOpenScience/exp-addons/commit/d30dd06273c9e9ffc96ae556aa3135e04f5f0188

exp-player/addon/components/exp-player.js
samchrisinger: Please feel free to change this implementation as you work through https://openscience.atlassian.net/browse/LEI-109

samchrisinger:
Closing in favor of https://github.com/CenterForOpenScience/exp-addons/pull/51

PR #33, opened 2016-03-02
Add the video preview frame

chrisseto:


exp-player/addon/components/exp-video-preview.js
```diff
@@ -0,0 +1,105 @@
...
+                },
+                hideRecorder: {
+                    type: 'boolean',
+                    description: 'Hide the video recorder component',
+                    default: true,
+                },
+                prompt: {
+                    type: 'string',
+                    description: 'Require a button press before showing the videos',
+                    default: false
```
abought: I'm not totally clear on what this variable is for- can you clarify?

The type is specified as a string, but the default is a boolean. (we get away with this mainly because the schema issn't used to validate anything; it's currently informational only)

It's also used as both a string and booleany in the corresponding template:
Button label: https://github.com/CenterForOpenScience/exp-addons/pull/33/files#diff-e31824100dedf21c87504a4b3ce9a7bdR14

Conditional: https://github.com/CenterForOpenScience/exp-addons/pull/33/files#diff-e31824100dedf21c87504a4b3ce9a7bdR6

abought:
Pausing review; will continue in AM.

PR #34, opened 2016-03-03
Support demo experiment

samchrisinger:
# Note

You must run `npm install` in the exp-player directory for this to work


exp-models/addon/models/experiment.js
```diff
@@ -25,11 +25,18 @@ export default DS.Model.extend(JamModel, {
     structure: DS.attr(),
     displayFullscreen: DS.attr('boolean'),
 
+    duration: DS.attr('string'),
```
abought: All of these newly added fields should be added to the experiment definition in `generate_schemas` before we forget.

exp-models/addon/models/experiment.js
```diff
@@ -25,11 +25,18 @@ export default DS.Model.extend(JamModel, {
...
+    whatHappens: DS.attr('string'),
+    purpose: DS.attr('string'),
```
abought: Are `whatHappens` and `purpose` separate from the existing `description` field? We may want to reorganize the schema and avoid making the fields too specific.

exp-models/addon/models/experiment.js
```diff
@@ -120,6 +127,7 @@ export default DS.Model.extend(JamModel, {
         var collection = this.store.createRecord('collection', {
             id: 'experimenter.' + this.get('sessionCollectionId')
```
abought: Instead of a hardcoded value, can you change this to use the `ENV.JAMDB.namespace` setting?

exp-player/addon/components/exp-exit-survey.js
```diff
@@ -148,7 +147,7 @@ export default ExpFrameBaseComponent.extend({
                 Ember.merge(formData, this.getValue());
                 root.set('formData', formData);
                 console.log('Post-study survey complete.');
-                root.actions.next();
+                root.actions.next.apply(root);
```
abought: Instead of `root.actions.next.apply`, use `root.send('actionName', arguments)`
http://emberjs.com/api/classes/Ember.Component.html#method_send

(Or `sendAction` for an action that was passed in and does not live in this controller)

exp-player/addon/components/exp-player.js
```diff
@@ -58,7 +58,10 @@ export default Ember.Component.extend(FullScreen, {
                 expData: this.get('expData'),
                 sequence: sequence
             };
-            this.sendAction('saveHandler', payload);  // call the passed-in action with payload
+            var exitUrl = this.get('experiment.exitUrl');
+            this.sendAction('saveHandler', payload,  () => {
```
abought: Can you please clarify what the third argument does? The documentation suggests there are only two parameters.
http://emberjs.com/api/classes/Ember.Component.html#method_sendAction

Since this is a somewhat route-related concern, you may wish for the window.location (or `this.transitionTo(urlish)`) call to not live in the player.

exp-player/addon/helpers/breaklines.js
```diff
@@ -0,0 +1,10 @@
+// h/t http://stackoverflow.com/questions/12331077/does-handlebars-js-replace-newline-characters-with-br
```
abought: Drawing from the Stackoverflow answer: 

What is the advantage of adding this helper rather than using a single [CSS style](https://developer.mozilla.org/en-US/docs/Web/CSS/white-space)?

exp-player/addon/mixins/full-screen.js
```diff
@@ -39,7 +39,7 @@ export default Ember.Mixin.create({
 
     actions: {
         showFullscreen: function (elementId) {
-            elementId = elementId || this.get('fullScreenElementId');
+            elementId = this.get('fullScreenElementId');
```
abought: If we're ignoring any value passed in, can I suggest removing the function parameter (now unused) and changing this line to `var elementId`?

exp-player/addon/templates/components/exp-audioplayer.hbs
```diff
@@ -1,13 +1,26 @@
...
+      </button>
     </div>
+  {{/unless}}
 
     <div class="col-md-10">
+      {{#if title}}
+        <h3>{{title}}</h3>
+      {{/if}}
+      {{#if titlePrompt}}
+        {{#with titlePrompt as |prompt|}}
```
abought: Does this really need to be aliased?

If we are aliasing this, it kind of suggests to me that the original names are confusing, and that may be the better place to make changes.

exp-player/addon/templates/components/exp-player.hbs
```diff
@@ -1,19 +1,16 @@
...
-        {{#if (and (not isFullscreen) experiment.displayFullscreen)}}
-            <button class="btn btn-warning pull-right" {{ action 'showFullscreen' }}>Return to fullscreen</button>
-        {{/if}}
-    </div>
+  session=session
+  saveHandler=(action 'saveFrame')
+  next=(action 'next') last=(action 'last') previous=(action 'previous')
+  }}
 </div>
+<div class="col-md-4 pull-right">
```
abought: Purely as a general note, most of our files are indented 4 spaces. (instead of 2)
https://cosdev.readthedocs.org/en/latest/style_guides/html_css.html

Not at all a blocker for merge, but watching for these things can save us some noisy formatting commits later.

exp-player/addon/templates/components/exp-video-setup.hbs
```diff
@@ -0,0 +1,30 @@
+<div class="row">
+  <h3> {{ title }} </h3>
```
abought: Informational only:

Some devices treat heading levels as having special meaning; try to avoid using non-sequential tags (h3-->h5) purely for formatting purposes. 
http://accessibility.psu.edu/headingshtml/

> Many tags in HTML, including the Heading tags (H1,H2,H3…H6) were developed not to assist formatting but to provide information on the structural hierarchy of a document... Many experts recommend reserving H1 for the page title, H2 for major headings and H3 for major sub headings.

exp-player/addon/components/exp-exit-survey.js
samchrisinger: I was having difficulty getting this to work without the strategy used here. Not sure why, but the action seemed to be skipping the exp-frame-base action.

If you can confirm a solution that is more Ember-friendly I'd be happy to change this, but I did spend ~20-30 min trying to figure this out and hesitate to invest too much more time.

exp-player/addon/components/exp-player.js
samchrisinger: It's a callback function. Looks like the corresponding code from https://github.com/CenterForOpenScience/experimenter/pull/55/files was lost during merge.  

Will update to a list before merge.

exp-player/addon/helpers/breaklines.js
samchrisinger: Have you confirmed that this works (in the context of a bound Ember attribute)? If so I will change it.

abought:
Review pass complete. I'll pick this up for functional review once the URLs work.

samchrisinger:
> merges PR in response to snark

abought:
> merges PR in response to snark

Er, that came out wrong. I just meant that the broken URLs (which may be a side effect of my stuff) are blocking me from actually trying out the cool new stuff.

exp-player/addon/helpers/breaklines.js
abought: Yes, I have confirmed that applying a CSS style to a span will suffice. There is no need to write handlebars code that rewrites or transforms the original text; let the browser handle presentation.

Here is the page if line 41 of audioplayer uses your helper:
screen shot 2016-03-04 at 11 27 34 pm

Compared to the page if line 41 says `<span style="white-space:pre-line">{{prompt.text}}</span>`:
screen shot 2016-03-04 at 11 31 56 pm

exp-player/addon/components/exp-exit-survey.js
abought: Interesting- it's because `formActions` aren't actions at all. It's a lightly documented hash of functions tied to buttons. (the guy behind dynamic forms gave a talk recently and acknowledged that he hadn't put full attention into documentation just yet)

That could be gnarly indeed; we can revisit this later when we have more time.
https://toddjordan.github.io/ember-cli-dynamic-forms/#/demos/actions

exp-models/addon/models/experiment.js
samchrisinger: Added in 4d61459f0b93ec8b95914a36019ad99491e184ac

exp-models/addon/models/experiment.js
samchrisinger: Added in 4d61459f0b93ec8b95914a36019ad99491e184ac

exp-models/addon/models/experiment.js
samchrisinger: Changed in 5345e50a31a9bacce1c4892401480fb1e717bd63

exp-player/addon/components/exp-player.js
samchrisinger: Changed in cfb6eb0e7927c4bb7281880e7ba768983014cb1f

samchrisinger:
This PR is blocking work on the Lookit site. I did my best to address all of the comments, and any needed changes can be made in a followup PR.

PR #36, opened 2016-03-03
[WIP] Feature/mood questionnaire

leb2dg:
### In Progress
* any time picker after the first shows up in a strange place

### Current State
screen shot 2016-03-08 at 2 55 51 pm

screen shot 2016-03-03 at 4 32 40 pm


exp-player/addon/components/exp-mood-questionnaire.js
```diff
@@ -0,0 +1,361 @@
...
+    }),
+    formData: [],
+    formActions: Ember.computed(function() {
+        var root = this;
+        return {
+            update: function () {
+                this.refreshValidationState(true);
+                if (this.isValid(true)) {
+                   root.set('formData', this.getValue());
+                   root.actions.next();
```
samchrisinger: This needs to be `root.actions.next.apply(root);` for this to work (the `this` context is wrong otherwise).

Also looks like these lines may be indented 3 spaces and not 4?

samchrisinger:
Some minor UI things:
 
screen shot 2016-03-07 at 3 57 54 pm

and

screen shot 2016-03-07 at 3 57 59 pm

could use a little more spacing/margins

PR #42, opened 2016-03-08
[WIP] Video recorder + viewer

chrisseto:


exp-player/addon/components/exp-player.js
```diff
@@ -41,52 +40,46 @@ export default Ember.Component.extend(FullScreen, {
...
-
             var frameIndex = this.get('frameIndex');
             if (frameIndex !== 0) {
+                console.log(`Previous: Transitioning to frame ${frameIndex - 1}`);
                 this.set('frameIndex', frameIndex - 1);
+            } else {
+                console.log('Previous: At frame 0');
             }
         },
         last() {
```
samchrisinger: We should trash this method. It was meant to allow skipping forward/backward to an arbitrary frame, but we don't have this use case.

chrisseto: Can do :+1:

PR #44, opened 2016-03-09
Expose experiment._sessionCollection

samchrisinger:
So consuming code can check if the session collection gets created properly

### DIFF
--- a/exp-models/addon/models/experiment.js
+++ b/exp-models/addon/models/experiment.js
@@ -122,6 +122,10 @@ export default DS.Model.extend(JamModel, {
         // Eg an experiment called 'test0' would have a collection 'session-test0'
         return `session${this.get('shortId')}s`; // FIXME: the spurious s is a requirement imposed by genschemas...
     }),
+    // Only use to check that the session collection is created when an
+    // experiment document is created. NOT a reliable way to fetch an
+    // experiment's session collection
+    _sessionCollection: null,
 
     _registerSessionModels() {
         // Dynamically register the required models for a session table associated with this experiment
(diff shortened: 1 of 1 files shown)


exp-models/addon/models/experiment.js
```diff
@@ -149,6 +153,12 @@ export default DS.Model.extend(JamModel, {
                 [`jam-${config.JAMDB.namespace}:accounts-*`]: 'CREATE'
             }
         });
-        collection.save();
+        this.set('_sessionCollection', collection);
+        collection.save().then(() => {
+            this.set('sessionCollectionId', this.get('sessionCollectionId'));
```
abought: `sessionCollectionId` is a computed, and does not depend on the newly created session. Why are you trying to set it?

exp-models/addon/models/experiment.js
```diff
@@ -122,6 +122,10 @@ export default DS.Model.extend(JamModel, {
         // Eg an experiment called 'test0' would have a collection 'session-test0'
         return `session${this.get('shortId')}s`; // FIXME: the spurious s is a requirement imposed by genschemas...
     }),
+    // Only use to check that the session collection is created when an
+    // experiment document is created. NOT a reliable way to fetch an
+    // experiment's session collection
+    _sessionCollection: null,
```
chrisseto: Any reason to not make this a computed?

samchrisinger: I'm not trying to encourage people to use this as a convenient way to fetch the experiment's sessions collection. This approach, while limited requires much less logic.

exp-models/addon/models/experiment.js
samchrisinger: Good point

PR #47, opened 2016-03-09
Do not allow login unless user has namespace permissions

abought:
Refs https://openscience.atlassian.net/browse/LEI-117
Companion to https://github.com/CenterForOpenScience/experimenter/pull/69

### DIFF
--- a/exp-models/addon/authenticators/jam-osf-jwt.js
+++ b/exp-models/addon/authenticators/jam-osf-jwt.js
@@ -1,19 +1,19 @@
 import Ember from 'ember';
-import Base from 'ember-simple-auth/authenticators/base';
+import BaseAuthenticator from 'ember-simple-auth/authenticators/base';
 
 
 import config from 'ember-get-config';
 
-export default Base.extend({
+export default BaseAuthenticator.extend({
     authUrl: `${config.JAMDB.url}/v1/auth`,
     namespaceUrl: `${config.JAMDB.url}/v1/id/namespaces/${config.JAMDB.namespace}`,
-    _get: function(accessToken) {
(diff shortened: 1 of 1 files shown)


exp-models/addon/authenticators/jam-osf-jwt.js
```diff
@@ -34,6 +48,14 @@ export default Base.extend({
         return this._post(accessToken).then(function(res) {
             res.data.attributes.accessToken = accessToken;
             return res.data.attributes;
+        })
+            .then((res) => {
```
chrisseto: Remove the extra new line here.
`}).then(res => ...`

exp-models/addon/authenticators/jam-osf-jwt.js
```diff
@@ -34,6 +48,14 @@ export default Base.extend({
...
+               // Only allow someone to login to the admin panel if they have have namespace-level permissions for this project
+                return this._askNamespace(res.token)
+                    .then(() => {return res;})
```
chrisseto: You don't need the `{}` or the `return` here.
`.then(() => res)` will do the same thing

exp-models/addon/authenticators/jam-osf-jwt.js
```diff
@@ -34,6 +48,14 @@ export default Base.extend({
...
+                    .fail(()=> {
```
chrisseto: See above.

Please add a space between `()` and `=>`

abought:
While I would have slightly preferred to not open the laptop tonight, careful reading deserves a timely response. :) All 3 requested changes pushed.

PR #48, opened 2016-03-09
Add mixin classes for notifiying a user when they try to leave a stud…

samchrisinger:
# Purpose

We need a way to warn users if they try to leave a study early that is generalizable to Experimenter and Lookit. Before we were checking the dirty state of the session and alerting the user if they tried to leave early. 

More realistically we want to alert the user no matter when they leave a study (for Lookit at least, a study is non-resumable). This implements this behavior as mixing classes and a component, and its use is demonstrated in https://github.com/CenterForOpenScience/experimenter/pull/70

### DIFF
--- a/exp-player/addon/templates/components/exit-warning.hbs
+++ b/exp-player/addon/templates/components/exit-warning.hbs
@@ -22,7 +22,7 @@
         </p>
         <div class="form-group">
           <div class="checkbox">
-            {{radio-button value="publicity and educatonal" checked=exitPrivacy}} Publicity and educational (video may be used for publicity or educa-onal purposes online or in the press in additon to for research purposes)
+            {{radio-button value="publicity and educatonal" checked=exitPrivacy}} Publicity and educational (video may be used for publicity or educational purposes online or in the press in additon to for research purposes)
           </div>
           <div class="checkbox">
             {{radio-button value="scientific" checked=exitPrivacy}} Scientific use only (e.g., a short clip may be shown at a scientific conference to demonstrate typical responses)
--- a/exp-player/addon/mixins/warn-on-exit-route.js
+++ b/exp-player/addon/mixins/warn-on-exit-route.js
@@ -6,7 +6,7 @@ export default Ember.Mixin.create({
             if(confirmed) {
                 controller.set('forceExit', true);
                 var session = this.get('controller.session');
-                session.set('didComplete', false);
+                session.set('complete', false);
                 session.set('earlyExit', data);
                 session.save();
                 transition.retry();


exp-player/addon/components/exit-warning.js
```diff
@@ -0,0 +1,20 @@
...
+import layout from '../templates/components/exit-warning';
+
+export default Ember.Component.extend({
+    layout,
+    warn: false,
+    exitReason: null,
+    exitPrivacy: null,
+    actions: {
+        onConfirm() {
+            this.get('onConfirm')({
```
chrisseto: Any reason to not just let the actions bubble up here? Or just call `this.send` or  `this.sendAction`

exp-player/addon/components/exp-video-consent.js
```diff
@@ -13,7 +13,7 @@ export default ExpFrameBaseComponent.extend({
                 this.get('videoRecorder').on('onUploadDone', null);
                 this.sendAction('next');
             });
-            this.get('videoRecorder').start(`video-consent-${this.get('session.id')}`, this.$().find('.recorder'), {
+            this.get('videoRecorder').start(`video-consent-${this.get('session.id')}`, Em.$(this.element).find('.recorder'), {
```
chrisseto: Why was this switched away from using `this.$`?

samchrisinger: It stopped working :(. I'll look into why

chrisseto: My PR for video play/recording has a fix for the video consent. I'm not sure if its the same issue.

exp-player/addon/mixins/warn-on-exit-controller.js
```diff
@@ -0,0 +1,33 @@
+import Ember from 'ember';
+
+export default Ember.Mixin.create(Ember.Evented, {
+    showWarning: false,
+    forceExit: false,
+
+    session: Ember.computed(function() {
+        return this.get('model.session') || this.get('model');
```
chrisseto: This is why I'm against Hashes

abought: That seems more like an issue with a mixin trying to fit into more than one use case, honestly. 

It might even be a good idea to specify the path to the session explicitly- if the mixin can't find the model it expects, guessing blindly might end with it checking the wrong model by accident.

(I agree that hashes get weird when the model hook is invoked in different ways, but that's a different issue)

exp-player/addon/templates/components/exit-warning.hbs
```diff
@@ -0,0 +1,55 @@
...
+      </div>
+    </div>
+    <hr>
+    <div class="row">
+      <div class="col-xs-12">
+        <div class="btn-group pull-right">
+          <button class="btn btn-default exit-warning-btn" {{action 'onReject'}}>
+            Back
+          </button>
+          <button disabled={{if (not exitPrivacy) true false}} class="btn btn-warning exit-warning-btn" {{action 'onConfirm'}}>
```
chrisseto: Can't this just be `disabled={{not exitPrivacy}}`?

samchrisinger: I think the truth helpers have to be inside of if/unless blocks

exp-player/addon/mixins/warn-on-exit-controller.js
samchrisinger: This will get simplified by https://github.com/CenterForOpenScience/exp-addons/pull/51, I.E. `session` will always be set on the controller.

The existing approach has the fallback mostly to cooperate with the lookit implementation

PR #49, opened 2016-03-09
Play/pause video playback on spacebar press

samchrisinger:


exp-player/addon/components/exp-video.js
```diff
@@ -11,6 +13,31 @@ export default ExpFrameBaseComponent.extend(FullScreen, MediaReload, {
     displayFullscreen: true,  // force fullscreen for all uses of this component, always
     fullScreenElementId: 'player-video',
 
+    spaceHandler: null,
+    didRender() {
```
chrisseto: Why not use `didInsertElement`

samchrisinger: Can you please explain or point me to a reference explaining why this might be better?

chrisseto: `didInsertElement` will only run once. This has the potential to run multiple times.

samchrisinger: If the component re-renders might it be a good idea to re-bind event listeners (especially with `player` closured in).

PR #50, opened 2016-03-09
Add completed field to session model

abought:
Refs https://openscience.atlassian.net/browse/LEI-133
Companion to https://github.com/CenterForOpenScience/experimenter/pull/73

### DIFF
--- a/exp-models/addon/models/session.js
+++ b/exp-models/addon/models/session.js
@@ -17,6 +17,8 @@ export default DS.Model.extend(JamModel, {
     experimentId: DS.attr('string'),
     experimentVersion: DS.attr('string'),  // TODO: Currently this field is not acted on in any way
 
+    completed: DS.attr('boolean'),  // Filter out sessions that were started, but never finished
+
     // Researchers can provide feedback to participants by writing to this field
     feedback: DS.attr('string'),
     // A flag for whether or not the participant has seen this feedback


chrisseto:
Looks like this might overlap with https://github.com/CenterForOpenScience/exp-addons/pull/48/files#diff-72d131960eb0a03351e599cc7d33a059R20

abought:
> Looks like this might overlap with https://github.com/CenterForOpenScience/exp-addons/pull/48/files#diff-72d131960eb0a03351e599cc7d33a059R20

Good catch. But while I like the way that other PR connects the attribute to a dirty exit check, I think this implementation has a few points to recommend it:
- Nomenclature: Ember uses the pattern `didVerb` for various model events, so I tried to avoid it in choosing an attribute name (`completed` vs `didComplete`)
- When to set a value: the other PR marks something as incomplete on exit, whereas this PR assumes something is incomplete until told otherwise. (seems a bit more fault-tolerant)
- Connections to other pages (the value is used to filter various pages where sessions are displayed)

Separately, the companion PR attempts to fix the issue where the preview page fails to load. (server rejects values of null being sent for empty values) This might be a useful change in itself if the other PR does not also address it.

Thoughts on implementation preference, @samchrisinger  ?

samchrisinger:
I think the dirty check is nice, but maybe impractical. For the Lookit use case at the very least, if the user leaves the study at any point (dirty or not), that session has ended. If the study was resumable, it would make complete sense to only warn if the state is dirty, but since in our system leaving the study is final I think the behavior should be a bit different.

> When to set a value: the other PR marks something as incomplete on exit, whereas this PR assumes something is incomplete until told otherwise. (seems a bit more fault-tolerant)

Can you please explain this thought ^ a bit more? Or maybe that's rolled into the first explanation? Maybe a little extra logic is needed to be sure experiments that exit normally don't get marked as incomplete?

PR #51, opened 2016-03-10
A better way of fetching past sessions?

samchrisinger:
Use a route mixin to make the exp-player route load until all of the
needed data is ready and successfully saved. This allows us to leverage
the loading substate of that route to give better feedback to the user,
and give better confidence that once the player is loaded it has
everything it needs to run.

### DIFF
--- a/exp-models/addon/models/account.js
+++ b/exp-models/addon/models/account.js
@@ -1,6 +1,7 @@
 /*
 Manage data about one or more documents in the accounts collection
  */
+import Ember from 'ember';
 import DS from 'ember-data';
 
 import JamModel from '../mixins/jam-model';
@@ -23,9 +24,10 @@ export default DS.Model.extend(JamModel, {
 
         return profiles.filter(getProfile)[0];
     },
-    pastSessionsFor(experiment) {
+    pastSessionsFor(experiment, profile) {
+        var profileId = Ember.get(profile, 'id') || '*';
         return this.get('store').queryRecord(experiment.get('sessionCollectionId'), {
-            q: `profileId:${this.get('id')}*`
+            q: `profileId:${this.get('id')}${profileId}`
         });
     }
 });
--- a/exp-models/addon/models/account.js
+++ b/exp-models/addon/models/account.js
@@ -1,7 +1,6 @@
 /*
 Manage data about one or more documents in the accounts collection
  */
-
 import DS from 'ember-data';
 
 import JamModel from '../mixins/jam-model';
@@ -23,5 +22,10 @@ export default DS.Model.extend(JamModel, {
         };
 
         return profiles.filter(getProfile)[0];
+    },
+    pastSessionsFor(experiment) {
+        return this.get('store').queryRecord(experiment.get('sessionCollectionId'), {
+            q: `profileId:${this.get('id')}*`
+        });
     }
 });
--- a/exp-models/addon/services/current-user.js
+++ b/exp-models/addon/services/current-user.js
@@ -1,0 +1,40 @@
+import Ember from 'ember';
+
+import config from 'ember-get-config';
+
+export default Ember.Service.extend({
+    store: Ember.inject.service(),
+    session: Ember.inject.service(),
+    account: null,
+
+    init() {
+        this._super(...arguments);
+
+        this.get('session').on('invalidationSucceeded', () => {
+            this.set('account', null);
+        });
+    },
+
+    getCurrentUser() {
+        return new Ember.RSVP.Promise((resolve, reject) => {
+            if(!this.get('session.isAuthenticated')) {
+                resolve(null);
+            }
+            else {
+                if (this.get('account')) {
+                    resolve(this.get('account'));
+                }
+
+                var data = this.get('session.data.authenticated');
+                var accountId = null;
+                if(data.provider === 'osf') {
+                    accountId = config.JAMDB.testAccount.id;
+                }
+                else if(data.provider === `${config.JAMDB.namespace}:accounts`) {
+                    accountId = data.id;
+                }
+                resolve(this.get('store').findRecord('account', `${config.JAMDB.namespace}.accounts.${accountId}`).catch(reject));
+            }
+        });
+    }
+});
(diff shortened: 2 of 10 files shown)


exp-models/addon/services/current-user.js
```diff
@@ -0,0 +1,40 @@
...
+            }
+            else {
+                if (this.get('account')) {
+                    resolve(this.get('account'));
+                }
+
+                var data = this.get('session.data.authenticated');
+                var accountId = null;
+                if(data.provider === 'osf') {
+                    accountId = config.JAMDB.testAccount.id;
```
abought: If we're going to embed what is basically a development setting inside production code, I propose one of the following:

- Modify the sample data so that `testAccount` explicitly refers to an administrator, or,
- Create a followup ticket that explicitly prevents administrators from creating new sessions (except perhaps when `--environment` is `development`)

exp-player/addon/mixins/exp-player-route.js
```diff
@@ -0,0 +1,43 @@
...
+            return this._getSession(params, experiment).then((session) => {
+
+                this.set('_experiment', experiment);
+                this.set('_session', session);
+
+                return experiment.getCurrentVersion().then(versionId => {
+                    session.set('experimentVersion', versionId);
+                    return session.save().then(() => {
+                        return this.get('currentUser').getCurrentUser((account) => {
+                            return account.pastSessionsFor(experiment);
```
abought: If the same account has multiple profiles, it seems that we would want to restrict the list of sessions by profile ID, not just account. 

I can help with sample queries for this purpose if needed...

exp-models/addon/services/current-user.js
```diff
@@ -0,0 +1,40 @@
...
+                }
+                else if(data.provider === `${config.JAMDB.namespace}:accounts`) {
+                    accountId = data.id;
+                }
+                resolve(this.get('store').findRecord('account', `${config.JAMDB.namespace}.accounts.${accountId}`).catch(reject));
```
abought: Is `currentUser` defined as a profile, or an account? (I imagine that login and experiment data collection may have different needs)

exp-models/addon/models/account.js
```diff
@@ -23,5 +22,10 @@ export default DS.Model.extend(JamModel, {
         };
 
         return profiles.filter(getProfile)[0];
+    },
+    pastSessionsFor(experiment) {
+        return this.get('store').queryRecord(experiment.get('sessionCollectionId'), {
+            q: `profileId:${this.get('id')}*`
```
abought: See notes elsewhere: should query by profile ID, not just account.

exp-models/addon/services/current-user.js
samchrisinger: @chrisseto is working on the latter approach

exp-models/addon/models/account.js
brianjgeiger: For Lookit, you rarely want to query by different profiles, but you often want to group by different profiles. But for any given route, I think you will almost always want all the past sessions for the whole account.

One example is the past studies page, where you're listing the previous sessions and the feedback. You'd have every previous session listed regardless of which child took the session. For the longitudinal studies, you'd want to be able to calculate which child is in which state, but you'd need to do that for every child at the same time.

That being said, I may be missing a detail of implementation, so feel free to ignore.

samchrisinger: I'm updating this to support a second optional parameter `profile`. If `profile` is unspecified it will use the wildcard, otherwise the query will be profile specific.

exp-player/addon/mixins/exp-player-route.js
```diff
@@ -0,0 +1,43 @@
...
+
+    },
+    _getSession(params, experiment) { // jshint ignore: line
+
+    },
+    model(params) {
+        // While a little gross, this ensures all the criteria for participation
+        // are met before the route resolves. This has the benefit that the route's
+        // loading state is active until all of this is complete.
+        return this._getExperiment(params).then((experiment) => {
```
abought: Lower priority than sorting out profiles vs accounts, but...

You may be able to simplify some of this using RSVP.hash (which doesn't resolve until all constituent promises are resolved, thus ensuring the model hook doesn't finish). 

While this blog's advice tends to be of very mixed quality, here is an example that might be useful. (incl using `setProperties`)
http://emberigniter.com/load-multiple-models-single-route/#rsvp-hash-to-the-rescue:7f3cf4d8f2178a468d49396e07502eab

samchrisinger: I definitely considered using a hash. There is some difficulty/complexity here with the dependent flow of data here (session needs experiment, pastSessions needs experiment + account + profile, etc).

abought:
Also, is there a JIRA ticket associated with this work? If so feel free to assign it to me for the review stage.

exp-models/addon/services/current-user.js
```diff
@@ -0,0 +1,54 @@
...
+
+        this.get('session').on('invalidationSucceeded', () => {
+            this.set('account', null);
+        });
+    },
+
+    getCurrentUser() {
+        return new Ember.RSVP.Promise((resolve, reject) => {
+            if(!this.get('session.isAuthenticated')) {
+                resolve(null);
```
chrisseto: Should this be a rejection?

samchrisinger: I kind of want to reserve that for error handling, but I could go either way

exp-models/addon/services/current-user.js
```diff
@@ -0,0 +1,54 @@
...
+                    profileId = ADMIN.profileId;
+                }
+                else if(data.provider === `${config.JAMDB.namespace}:accounts`) {
+                    accountId = data.id;
+                    profileId = this.get('data.profile.profileId');
+                }
+                resolve(
+                    this.get('store').findRecord('account', `${config.JAMDB.namespace}.accounts.${accountId}`)
+                        .then((account) => {
+                            return [account, account.profileById(profileId)];
```
chrisseto: I'd prefer that resolve get called in here rather than calling it on this entire promise.

chrisseto: And there is no need for the `return` or `{}`s as this is a single line lambda function.

exp-models/addon/services/current-user.js
chrisseto: Fair, it's weird to juggle a success, failed, and kinda failed state.

PR #61, opened 2016-03-14
Anonymize profile IDs displayed in result summary table

abought:
Refs: https://openscience.atlassian.net/browse/LEI-149
Companion to: https://github.com/CenterForOpenScience/experimenter/pull/84

### DIFF
--- a/exp-models/addon/models/session.js
+++ b/exp-models/addon/models/session.js
@@ -37,6 +37,11 @@ export default DS.Model.extend(JamModel, {
 
     }),
 
+    anonProfileId: Ember.computed('profileId', function() {
+        // For a profile id of format `<acctShortId>.random`, strip off the identifying account ID
+        var profileId = this.get('profileId');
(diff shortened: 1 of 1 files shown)


exp-models/addon/models/session.js
```diff
@@ -37,6 +37,11 @@ export default DS.Model.extend(JamModel, {
 
     }),
 
+    anonProfileId: Ember.computed('profileId', function() {
+        // For a profile id of format `<acctShortId>.random`, strip off the identifying account ID
+        var profileId = this.get('profileId');
+        return profileId.split ? profileId.split('.')[1] : profileId;}),
```
chrisseto: When will split not be defined?

abought: The main use case I had in mind was if the profile ID was of some other type (eg non-numeric).

Although this will still break if `profileId` is undefined, that shouldn't happen; any session without a defined profile ID would be fairly seriously borked in other ways.

PR #65, opened 2016-03-15
Support Lookit + revised URLs

abought:
Refs: https://openscience.atlassian.net/browse/LEI-160
Companion to: https://github.com/CenterForOpenScience/lookit/pull/44

## Purpose
Fix small issues with lookit + newest codebase.

## Testing notes
This should fix adapter errors that occur when trying to participate in an experiment via lookit. (due to the current user service requesting `experimenter.accounts.experimenter.accounts.tester0` or such).

At the end of the experiment, you should now be able to opt out of emails, and have data submit to server. Previously, due to the lack of a default value, this resulted in errors of "Uncaught TypeError: Cannot set property 'emailOptOut' of undefined" being dumped to the console. (which prevented the value from being sent)

### DIFF
--- a/exp-player/addon/components/exp-frame-base.js
+++ b/exp-player/addon/components/exp-frame-base.js
@@ -65,10 +65,11 @@ export default Ember.Component.extend({
         });
 
         Object.keys(this.get('meta.data').properties || {}).forEach((key) => {
-            if (this[key] && this[key].isDescriptor) return;
+            if (this[key] && this[key].isDescriptor) {return;}
             var value = !clean ? this.get(key): undefined;
             if (typeof value === 'undefined') {
-                defaultParams[key] =  this.get(`meta.data.properties.${key}.default`);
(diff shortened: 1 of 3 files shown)


exp-player/addon/components/exp-frame-base.js
```diff
@@ -65,10 +65,11 @@ export default Ember.Component.extend({
         });
 
         Object.keys(this.get('meta.data').properties || {}).forEach((key) => {
-            if (this[key] && this[key].isDescriptor) return;
```
chrisseto: There's no need to add the curly braces here. Please remove them.

abought: I don't have a strong inherent preference, but this change was made to reduce the number of Jshint errors, so that linting remained useful and informative.

This setting can be overridden (`curly: false`), since the preference is a matter of opinion... but on average I think there are fewer bugs when we enforce the convention globally than when we don't. 

(and using `//jshint ignore: line` to carve out an exception in this case would just be gratuitous)

PR #70, opened 2016-05-19
Pref phys addons

kimberscott:
Here are my changes so far in tweaking the first three frames of the "Preferential  Physics" study on experimenter. 

On the exp-physics-preview frame, you have to click twice on the prompt button for the video to show up. (Each time, accept is called.) I though this might just be about the prompt (maybe some check for prompt was happening before it was updated) but if you don't use the prompt, you instead have to click "next video" twice. (Each time, nextVideo is called and actually does appear to increment the index--from 0 to 1 each time.)


samchrisinger:
The double-clicking issue should be fixed in https://github.com/CenterForOpenScience/exp-addons/commit/17f11695c68d58f9123a1821a864f2946a787517. Seems like the `index` property was conflicting with another usage.

PR #72, opened 2016-05-24
Pref phys addons

kimberscott:
Current progress on preview & mood questionnaire frames. Works w/ "Preferential physics" study on experimenter.


kimberscott:
Intro and dummy test videos are up on S3 and this pref-phys-video section does actually run now! Next steps are:

- getting fullscreen working (currently only the first frame stays fullscreen) and then setting appropriate placement/sizing of videos

- adding pause functionality (basic pausing/restarting; having pause -> show attentiongrabber video; restart at intro with either the same test video or a flipped version depending on when pausing happened)

- adding verbal instructions for the parent and music

- put up the real videos, or as close as we have

exp-player/addon/components/exp-video-physics.js
```diff
@@ -0,0 +1,172 @@
...
+
+    didRender() { // set up event handler for pausing
+        this._super(...arguments);
+        var emberObj = this;
+        $(document).on("keypress", function (e) {
+            if (emberObj.checkFullscreen()) {
+                if (e.which == 32) { // space
+                    emberObj.send('pause');
+                    // This isn't actually a function but somehow triggering the error works anyway so I'll take it for now...
+                    //emberObj.get('refresh')();
```
samchrisinger: @kimberscott can you explain a bit about what's going on here? What's the intended behavior when calling `emberObj.refresh()`?

samchrisinger:
I'm going to go ahead and merge this so as to not block progress and will address questions as a followup.

exp-player/addon/components/exp-video-physics.js
kimberscott: This was part of some debugging to figure out why the handler was consistently only being called the second time the event happened. The intended behavior was probably to try forcing re-rendering, maybe in case pause was being called but we didn't see the changes? And it works--consistently, on both Firefox and Chrome. (The same thing "fixes" the double-click bug in preview...) 

But of course this isn't actually a valid function call. It also works with a made-up function name. So somehow the error is enough to fix this "double-click" (/"double-press") bug.  Possibly it's just some delay that's being introduced between key events?

exp-player/addon/components/exp-video-physics.js
```diff
@@ -0,0 +1,172 @@
...
+                    emb.send('playNext');
+                    }, emberObj.get('testLength')*1000, emberObj);
+                this.set('timeoutID', t);
+            }
+        },
+        pause: function () {
+            var movie = $('#player-video')[0];
+            var emberObj = this;
+            window.clearTimeout(this.get('timeoutID'));
+            if (!emberObj.get('doingAttn') || !this.checkFullscreen()) { // pausing one of the videos
```
samchrisinger: Are `doingAttn` and `doingIntro` mutually exclusive conditions? I read this logic like

> If not doing an attention grabber or not fullscreen, show a fullscreen attention grabber. Otherwise if not doing an intro video check if an alternate video is available. If so, show it, otherwise move move. Regardless hide the attention grabber.

I think we may be seeing competing 'pause' events from `onFullScreen` and the event handler for spacebar events. I started tweaking the code but am realizing I don't know exactly how this needs to work.

kimberscott:
`doingAttn` and `doingIntro` aren't mutually exclusive. 

`doingAttn`: are we currently showing the attentiongrabber?

`doingIntro`: are we currently in the "intro" phase of this trial? This is true both when the intro video is actually playing and if we paused the intro and are showing the attentiongrabber.

The logic is:

> 
> If we're currently playing a "real" video right now (not doingAttn), or if the window isn't fullscreen, show the attentiongrabber. (Fullscreen isn't altered. If someone hits space outside of fullscreen, nothing happens--we want to force them to return to fullscreen before continuing.)
> 
> Otherwise--i.e., if we ARE playing a "real" video right now AND the window is fullscreen--return to the intro video mode. If we were showing the intro before this introduction, we just return to it, no harm done. If we were already showing the alternate test video, move along to the next trial. Otherwise, set useAlternate to true so that when we complete this intro and move to the test clip, we'll show the alternate.

Higher-level take: babies should see a sequence of trials, each of which has an intro video followed by a test video. Sometimes their parents may need to turn around and pause the study, or some event on the computer may make the window leave fullscreen mode. When this happens, we'd like to pause video playback, and then when they're ready to continue, restart from the trial they were on. But if the study was paused during the test video, that video is "contaminated"--the parent may know which direction is the more interesting one, etc. So we show a slightly different video that tests the same thing instead. We only do up to one substitute per trial, though, so that pausing repeatedly can't make the study drag on forever (and because it would make the data less interpretable).

The actual logic is working correctly as is (with the dumb throw-an-error fix.)

exp-player/addon/components/exp-video-physics.js
samchrisinger: So I believe the issue with double-clicking was caused by multiple event handlers being attached here. Basically `didRender` gets run every time the component re-renders, which in this case was happening when the video sources were being swapped out. Changing this to `init` limits the event binding only to the first render.

Proposed fix here: https://github.com/CenterForOpenScience/exp-addons/pull/73/files#diff-848af9bc2447ff0e1ce2bc717ba79035R173

samchrisinger:
@kimberscott thanks for the explanation! I tried not to change the logic at all, but can you please double-check the changes made here: https://github.com/CenterForOpenScience/exp-addons/pull/73/files#diff-848af9bc2447ff0e1ce2bc717ba79035R173 ?

kimberscott:
Sure, just tried it out! Three issues:

- When leaving FS, the "return to FS" button isn't shown, so the user is stuck.

- If we get to the next exp-physics-video frame by way of exhausting the test and alternate video (i.e., if the user pauses during the test video, and then again during the alternate test video) then leaving FS does show the button but doesn't pause the video.

- The test videos are only being played once--they should loop for a total duration of testLength s.

PR #73, opened 2016-06-02
Refactor and tweaks to exp-video-physics

samchrisinger:


exp-player/addon/components/exp-video-physics.js
```diff
@@ -105,68 +127,55 @@ export default ExpFrameBaseComponent.extend(FullScreen, MediaReload, VideoPause,
...
-                    .attr('loop', true)
-                    .addClass('test')
-                    .removeClass('intro attn');
             } else {
-                this.send('next');
-                $('#player-video')
-                    .removeClass('attn test')
-                    .addClass('intro')
-                    .attr('loop', false);
+                this.get('next')(); // moving to intro video
```
samchrisinger: TODO: @samchrisinger changeme

kimberscott:
(Copying notes here because this is probably the more logical location) - issues:

- When leaving FS, the "return to FS" button isn't shown, so the user is stuck.

- If we get to the next exp-physics-video frame by way of exhausting the test and alternate video (i.e., if the user pauses during the test video, and then again during the alternate test video) then leaving FS does show the button but doesn't pause the video.

- The test videos are only being played once--they should loop for a total duration of `testLength` s.

kimberscott:
Looks great!

PR #80, opened 2016-06-08
Recording during exp-video-preview

kimberscott:
Adding a PR here to test video recording on the staging server. I followed the video frame example.


kimberscott:
Video recording during exp-video-preview is tested, and I've added some additional logging of timing events during the video sections.

PR #100, opened 2016-07-29
Feature: ISP consent form

samanehsan:
## Purpose
Add exp-frame for ISP consent from

## Summary of changes
- Make `exp-consent` form button text a variable 
- Add `exp-isp-consent` exp-frame that extends `exp-consent.js` and defines 
  additional properties
- `exp-isp-consent/template.hbs` includes translation helpers


samchrisinger:
Nice! Happy to see the consent form code reused.

PR #105, opened 2016-08-17
Add jam-auth code to exp-addons

samanehsan:
Add jam-auth component, authenticator & authorizer to exp-addons since it is used by both lookit and ISP

### DIFF
--- a/exp-models/addon/authenticators/jam-jwt.js
+++ b/exp-models/addon/authenticators/jam-jwt.js
@@ -1,0 +1,46 @@
+import Ember from 'ember';
+import config from 'ember-get-config';
+import BaseAuthenticator from 'ember-simple-auth/authenticators/base';
+
+import { jwt_decode } from 'ember-cli-jwt-decode';
+
+import moment from 'moment';
+
+const {
+    $, RSVP
+} = Ember;
+
+export default BaseAuthenticator.extend({
+  url: `${config.JAMDB.url}/v1/auth`,
+  restore(data) {
+    let token = JSON.parse(atob(data.token.split('.')[1]));
+    if (token.exp > moment().unix()) {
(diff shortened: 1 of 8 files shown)


exp-models/addon/authenticators/jam-jwt.js
```diff
@@ -0,0 +1,46 @@
...
+
+import { jwt_decode } from 'ember-cli-jwt-decode';
+
+import moment from 'moment';
+
+const {
+    $, RSVP
+} = Ember;
+
+export default BaseAuthenticator.extend({
```
samchrisinger: @samanehsan can you please reformat this code so it matches with https://github.com/CenterForOpenScience/exp-addons/blob/develop/exp-models/.editorconfig and the other code in this repo?

exp-player/addon/components/jam-auth/component.js
```diff
@@ -0,0 +1,25 @@
+import Ember from 'ember';
+import ENV from 'isp/config/environment';
```
samchrisinger: I think you'll need to use the `ember-get-config` addon here

exp-player/addon/components/jam-auth/template.hbs
```diff
@@ -0,0 +1,20 @@
+<div class="well">
+    <p>Enter your <strong>Study ID</strong> and <strong>Participant ID</strong> into the fields below. Please {{link-to 'contact us' 'contact'}} if you have any questions.</p>
```
samchrisinger: I would prefer to see the ISP-specific representation of this component included in that repo instead. 

Can you please replace this with a more generic version? The ISP version can create a subclass of this component and override `layout` to have this template.

samanehsan:
Done making those changes @samchrisinger!

PR #117, opened 2016-08-29
Feature/wireframe updates

samanehsan:
## Summary of changes
Free-response form: 
- Fix character count to show number of characters used
- Do not translate times, instead change it to say "19:00 (7pm)" instead of just "7pm"

Card-sort form:

- Add labels to the free-response answers displayed above the cards to clarify what question they belong to:  
screen shot 2016-08-29 at 4 04 43 pm

Rating form:  
 
- Alternate condition in rating-form question

## Testing notes
Needs https://github.com/CenterForOpenScience/isp/pull/44 merged

### DIFF
--- a/exp-player/addon/components/exp-free-response/component.js
+++ b/exp-player/addon/components/exp-free-response/component.js
@@ -28,11 +28,10 @@ export default ExpFrameBaseComponent.extend(Validations, {
     layout: layout,
     i18n: Ember.inject.service(),
     displayTime: Ember.computed(function() {
-        var profileId = this.get('session').get('profileId').split('.')[1];
-        var participantId = profileId.split('_')[1];
-        if (participantId % 2 === 0) {
+        var condition = this.get('session').get('experimentCondition');
+        if (condition === '7pm') {
             return '19:00 (7pm)';
-        } else {
+        } else if (condition === '10am') {
             return '10am';
         }
     }),
--- a/exp-player/addon/components/exp-free-response/component.js
+++ b/exp-player/addon/components/exp-free-response/component.js
@@ -32,9 +32,9 @@ export default ExpFrameBaseComponent.extend(Validations, {
         var profileId = this.get('session').get('profileId').split('.')[1];
         var participantId = profileId.split('_')[1];
         if (participantId % 2 === 0) {
-            return this.get('i18n').t('survey.sections.2.times.7pm').string;
+            return '19:00 (7pm)';
         } else {
-            return this.get('i18n').t('survey.sections.2.times.10am').string;
+            return '10am';
         }
     }),
     instructions: Ember.computed(function() {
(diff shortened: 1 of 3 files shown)


exp-player/addon/components/exp-free-response/component.js
```diff
@@ -48,21 +47,21 @@ export default ExpFrameBaseComponent.extend(Validations, {
       return label;
     }),
     diff1: Ember.computed('q1', function() {
-        var remaining = getRemaining(this.get('q1'));
+        var length = getLength(this.get('q1'));
```
samchrisinger: Might try `Ember.get('q1.length')` instead of a custom helper.

exp-player/addon/components/exp-rating-form/component.js
```diff
@@ -201,11 +201,11 @@ var generateValidators = function(questions) {
...
   });
-  for (var question in questions) {
-    for (var item in questions[question]['items']) {
-      var isOptional = questions[question]['items'][item].optional;
+  for (var i=0; i < questions.length; i++) {
+    for (var item=0; item < questions[i]['items'].length; item++) {
+      var isOptional = questions[i]['items'][item].optional;
       if (!isOptional) {
-        var key = 'questions.' + question + '.items.' + item + '.value';
+        var key = 'questions.' + i + '.items.' + item + '.value';
```
samchrisinger: Nothing wrong here, but remember you have access to es6 string templates, e.g.:
```var key = `questions.${i}.items.${item}.value`;```

File: exp-player/addon/components/exp-rating-form/component.js [hunk 3]
```diff
@@ -552,7 +551,14 @@ const Validations = buildValidations(generateValidators(questions));
 export default ExpFrameBaseComponent.extend(Validations, {
     type: 'exp-rating-form',
     layout: layout,
-    questions: questions,
+    questions: Ember.computed(function() {
+        var profileId = this.get('session').get('profileId').split('.')[1];
+        var participantId = profileId.split('_')[1];
+        if (participantId % 2 === 0) {
```
samchrisinger: Preferable to grab this value from the session rather than recomputing. If we every change how conditions are assigned it would be nice to know only one place in the code needs updating.

exp-player/addon/components/exp-free-response/component.js
samanehsan: The default value is `undefined` so it will complain if I try to do something like `Ember.get(this.get('q1'), 'length')`.

samanehsan:
Done making those changes @samchrisinger! ☀️  https://github.com/CenterForOpenScience/isp/pull/45 saves the experiment condition on the session so it needs to be merged before this one.

exp-player/addon/components/exp-rating-form/component.js
```diff
@@ -296,7 +295,7 @@ var questions = [
     }
   ),
   generateSchema(
-    'measures.questions.4.label',
+    'measures.questions.4.label.10am',
```
samchrisinger: Does this need to be dynamic based on the condition?

samanehsan: Yes, I determine that here: https://github.com/CenterForOpenScience/exp-addons/pull/117/files#diff-9150b292474824ef67f55d3899fd974aR556

I specified a default so I could generate validators for that question. It's based on the index of the question in the list, so the exact label doesn't matter for validation.

samchrisinger:
Thanks!

PR #124, opened 2016-09-08
Remove unused exp-isp-consent component

samanehsan:
## Summary of changes
Remove the exp-isp-consent component since the consent form is no longer part of the exp-frame sequence and a separate isp-consent component was added to the ISP code.


samchrisinger:
Please use the ember cli's `ember d component exp-isp-consent` so unit tests get deleted too.

samanehsan:
@samchrisinger, I must have not committed any unit tests to begin with because they're not there!

PR #127, opened 2016-09-12
Hide optional overview form question [LEI-192] 

samanehsan:
Purpose
-------
On the exp-overview form, if a user answers yes to question 8: "Do you follow a religion"
show question 9: "If so, which one do you follow?" Otherwise, hide it.

Changes
-------
Add a computed property `showOptional` that watches the value of question 8. Then in the template, only show that question if `showOptional` is true.

### DIFF
--- a/exp-player/addon/components/exp-overview/component.js
+++ b/exp-player/addon/components/exp-overview/component.js
@@ -117,7 +117,7 @@ export default ExpFrameBaseComponent.extend(Validations, {
         responses[i] = questions[i].value;
       }
       return responses;
-    }),
+    }).volatile(),
     meta: {
       name: 'ExpOverview',
         description: 'TODO: a description of this frame goes here.',
--- a/exp-player/addon/components/exp-overview/component.js
+++ b/exp-player/addon/components/exp-overview/component.js
@@ -107,13 +107,16 @@ export default ExpFrameBaseComponent.extend(Validations, {
     type: 'exp-overview',
     layout: layout,
     questions: questions,
-    responses: Ember.computed('questions', function() {
-        var questions = this.get('questions');
-        var responses = {};
-        for (var i=0; i < questions.length; i++) {
-            responses[i] = questions[i].value;
-        }
-        return responses;
+    showOptional: Ember.computed('questions.7.value', function() {
+       return this.questions[7].value === 'global.yesLabel';
+    }),
+    responses: Ember.computed(function() {
+      var questions = this.get('questions');
+      var responses = {};
+      for (var i=0; i < questions.length; i++) {
+        responses[i] = questions[i].value;
+      }
+      return responses;
     }),
     meta: {
       name: 'ExpOverview',
--- a/exp-player/addon/components/exp-overview/template.hbs
+++ b/exp-player/addon/components/exp-overview/template.hbs
@@ -3,14 +3,19 @@
     {{#bs-form}}
       {{#each questions as |question|}}
         {{#bs-form-group}}
-          <label>{{t question.question}}</label>
-          {{#if (eq question.type 'radio')}}
-            {{isp-radio-group options=question.scale labelTop=question.labelTop value=question.value}}
-          {{else if (eq question.type 'select')}}
-            {{select-input options=question.scale value=question.value}}
-          {{else if (eq question.type 'input')}}
-            {{#input class="form-control auto-width" value=question.value}}{{/input}}
-          {{/if}}
+          {{#unless question.optional}}
+            <label>{{t question.question}}</label>
+            {{#if (eq question.type 'radio')}}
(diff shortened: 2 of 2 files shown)


exp-player/addon/components/exp-overview/component.js
```diff
@@ -107,13 +107,16 @@ export default ExpFrameBaseComponent.extend(Validations, {
...
     layout: layout,
     questions: questions,
-    responses: Ember.computed('questions', function() {
-        var questions = this.get('questions');
-        var responses = {};
-        for (var i=0; i < questions.length; i++) {
-            responses[i] = questions[i].value;
-        }
-        return responses;
+    showOptional: Ember.computed('questions.7.value', function() {
```
samchrisinger: I don't think you can register a computed on a single array element like this. Probably your best bet is `questions.[].value`

exp-player/addon/components/exp-overview/component.js
```diff
@@ -107,13 +107,16 @@ export default ExpFrameBaseComponent.extend(Validations, {
...
+       return this.questions[7].value === 'global.yesLabel';
+    }),
+    responses: Ember.computed(function() {
```
samchrisinger: Should this computed depend on `questions`?

exp-player/addon/components/exp-overview/component.js
samanehsan: Unfortunately the only way to watch the `value` property nested under `questions[7]` is with `'questions.7.value'`.

samchrisinger: As long as this works I'm fine with it-- I've just never seen an example of registering a computed on a specific array element like this. 

Looking at https://guides.emberjs.com/v2.6.0/object-model/computed-properties-and-aggregate-data/, I don't see this documented anywhere

exp-player/addon/components/exp-overview/component.js
samanehsan: Watching `questions` would just watch for changes to the list itself, instead of changes to the `value` of each question. It looks like `responses` gets re-computed when you click `continue` so all of the answers are there.

samchrisinger: If you're counting on the computed being reevaluated when pressing 'continue' , I'd suggest adding `.volatile()` to the computed definition (http://emberjs.com/api/classes/Ember.ComputedProperty.html#method_volatile)

Or looking at the first example here: https://guides.emberjs.com/v2.6.0/object-model/computed-properties-and-aggregate-data/

you might be able to use `@each.value`

exp-player/addon/components/exp-overview/component.js
samanehsan: Oh, I think I was confusing lists with objects -- `questions.@each.value` would also work, but that would watch all values instead of just `question[7]`

samchrisinger: I'm not sure if Ember will allow you to observe single array elements, so this may be your only option.

exp-player/addon/components/exp-overview/component.js
samanehsan: For some reason, watching `'questions.@each.value'` works for `showOptional`, but not for `responses`. Does it have to do with `responses` being serialized?

samanehsan:
Done making that change @samchrisinger! 🚀

exp-player/addon/components/exp-overview/component.js
samchrisinger: @samanehsan still blocked on this comment

samanehsan: I thought the consensus was to leave it as `questions.7.value`. Did you want me to use `questions.@each.value` instead? Either of them will work.

exp-player/addon/components/exp-overview/component.js
abought: Incidentally, I'm using `questions.n.value` in an ember.alias, and it appears to be tracking the change so far. 
https://github.com/CenterForOpenScience/exp-addons/pull/131/files#diff-20e4394de68cd99661ed9d994056e5a6R112

However, I agree that the lack of documentation around it makes me nervous, and I'm watching for edge cases as I build out around this.

exp-player/addon/components/exp-overview/component.js
samchrisinger: Let's use the more specific `questions.7.value`. Looks like this works because the native JS Array will accept string index values:

> It's possible to quote the JavaScript array indexes as well (e.g., years['2'] instead of years[2]), although it's not necessary. The 2 in years[2] is coerced into a string by the JavaScript engine through an implicit toString conversion.

([which is how the Computed.get implementation works](https://github.com/emberjs/ember.js/blob/lts-2-4/packages/ember-metal/lib/property_get.js#L145))

PR #133, opened 2016-09-14
[WIP] Handle browser's back button [LEI-186]

samanehsan:
## Purpose
If a user clicks the back button, they will get redirected to the previous route instead of the previous exp-frame or frame page.

## Summary of changes
- Pass `frameIndex` and `framePage` to `exp-player`
- When a user is about to transition from the `survey/index` route, transition the user back to the `index` route and update the `frameIndex` and `framePage` to show the correct page.


exp-player/addon/components/exp-player.js
```diff
@@ -18,6 +18,7 @@ export default Ember.Component.extend(FullScreen, {
     conditions: null,
 
     frameIndex: 0, // Index of the currently active frame
+    framePage: 1, // Index of the currently visible page within a frame
```
samchrisinger: If we're going to add this to the exp-player we should unify how we're indexing `framePage` with `frameIndex` (both should be zero indexed)

exp-player/addon/components/exp-player.js
```diff
@@ -142,6 +143,7 @@ able to continue the study.
         next() {
             console.log('next');
             var frameIndex = this.get('frameIndex');
+            this.sendAction('updateFrameIndex', frameIndex);
```
samchrisinger: Does this work (I see no action handler for this action)? This should get handled in `_transition`

exp-player/addon/components/exp-player.js
```diff
@@ -181,6 +183,10 @@ able to continue the study.
         },
         closeExitWarning() {
             this.set('hasAttemptedExit', false);
+        },
+        updateFramePage(framePage) {
+            this.set('framePage', framePage);
+            this.sendAction('updateFramePage', framePage);
```
samchrisinger: For compatibility with existing usages (Lookit in particular) It would be ideal to check if the `updateFramePage` action was passed down or not (and only call this if the handler is available)

exp-player/addon/components/exp-rating-form/component.js
```diff
@@ -605,9 +605,17 @@ export default ExpFrameBaseComponent.extend(Validations, {
...
         nextPage() {
-            this.set('currentPage', this.currentPage + 1);
+            var page = this.framePage + 1;
+            this.set('framePage', page);
+            this.sendAction('updateFramePage', page);
             this.send('save');
             window.scrollTo(0,0);
+        },
+        previousPage() {
+            var page = this.framePage - 1;
```
samchrisinger: Has to be `this.get('framePage')

samchrisinger: Should this be `page = Math.min(0, this.get('framePage') - 1)`?

exp-player/addon/components/exp-player.js
samanehsan: This action gets bubbled up to the `participate.survey.index` controller, which handles the action: https://github.com/samanehsan/isp/blob/6a470e32d4101b6f3058b6123e8ee835d9510c8e/app/controllers/participate/survey/index.js#L19

samchrisinger: Does this create any redundancy with L150 (will frameIndex get incremented twice?)?

samanehsan: I don't think so, since it is only called in the `next` action.

samchrisinger: Sorry if I'm misreading or missing something... but won't the call on L146 (at least in ISP) increment `frameIndex`? The it gets incremented again here on L150.

samanehsan: Oh, I see what you mean -- the `updateFrameIndex` action updates the frameIndex property of the `participate.survey.index` controller and `this.set('frameIndex', frameIndex + 1);` on L150 is incrementing the exp-player's `frameIndex` property.

samanehsan: I'm tracking `frameIndex` and `framePage` on the `participate.survey.index` so that I know what frame/page the user is on when they hit the back button

samchrisinger: Yes, that same `frameIndex` is passed into the exp-player via the template [here](https://github.com/CenterForOpenScience/isp/pull/64/files#diff-2f922b8396c42d2f74d61c16c2624b7fR10). When the value is updated on the controller it will also get updated on the component.

samanehsan: Oh, so now that `frameIndex` is passed in, I can get rid of L150?

samchrisinger: That will break other usages of `exp-player`.

In some ways `exp-player` deviates from proper data-down-actions-up paradigms since it mutates `frameIndex` internally rather than calling an action on the parent controller/component. Since `frameIndex` is a bound property, I'd expect in your case that the controller's value would update too (without needing to send the `updateFrameIndex` action). But I'm guessing you tried this and found that wasn't the case?

On approach would be to generalize this behavior into an `exp-player-controller` mixin like the [`exp-route-mixin`](https://github.com/CenterForOpenScience/exp-addons/blob/develop/exp-player/addon/mixins/exp-player-route.js) that exposes actions for changing the `frameIndex` outside of exp-player. Might be worth looking into, but first can you confirm that `frameIndex` isn't already getting changed on the controller (without the new action)?

exp-player/addon/components/exp-player.js
samanehsan: Since `framePage` is passed in just like `frameIndex`and should update when the component's value changes, can I remove sending the action?

samanehsan:
Done making those changes @samchrisinger! ☀️

PR #136, opened 2016-09-16
[HOLD] Add ISP form validation back in [LEI-183]

samanehsan:
## Purpose
Do not let users go to the next section of the survey unless they have answered all of the questions on the page.

## Summary of changes
- This reverts commit 0354365e1e29d8066c9df77808a142f49e09f449.

## Note
Merge after the go-ahead from UCR team, since they requested removing validation until front-end testing was complete.


exp-player/addon/components/exp-rating-form/component.js
```diff
@@ -197,16 +197,19 @@ const NINE_POINT_SCALE = TEN_POINT_SCALE.slice(0, 10);
...
-    });
     for (var i = 0; i < questions.length; i++) {
         for (var item = 0; item < questions[i]['items'].length; item++) {
             var isOptional = questions[i]['items'][item].optional;
             if (!isOptional) {
                 var key = `questions.${i}.items.${item}.value`;
-                validators[key] = presence;
+                validators[key] = validator('presence', {
+                    presence: true,
+                    message: 'This field is required',
```
abought: Not now, but a note for the future: will validation text need to be translated?

abought:
Not a blocker now, but in the future, we may consider making it easier to turn validation on or off.

For example, each frame could have a computed property (`allowNext`) which responds to either a feature flag (in config/environment.js) or the validation state of the form.  This would facilitate testing (turning off validation) without needing to remove code manually.

exp-player/addon/components/exp-rating-form/component.js
samanehsan: This message actually doesn't get shown, since the continue button is disabled until all required questions have been answered.

PR #131, opened 2016-09-13
[WIP] Simplify data that gets sent to the server

abought:
Refs: https://openscience.atlassian.net/browse/LEI-193
Companion to: TBD

## Purpose
Make it easier to read serialized data, by removing extraneous information and simplifying the deep nesting.

When done, this will be a companion to an **experimenter** PR, which will provide a new serialization mechanism based on the simplified payload.

## Summary of changes

For the card sort responses section of the payload:

Before
```
                   "cardSortResponse": [{
                        "categories": [
                                {
                            "cards": [{
                                "isDraggingObject": true,
                                "id": "anxietyInducing",
                                "content": "qsort.rsq.item.anxietyInducing"
                            }, {
                                "isDraggingObject": true,
                                "id": "socialInteraction",
                                "content": "qsort.rsq.item.socialInteraction"
                            }, {
                                "isDraggingObject": true,
                                "id": "rapidlyChanging",
                                "content": "qsort.rsq.item.rapidlyChanging"
                            }], "max": 3, "name": "qsort.sections.2.categories.extremelyUnchar"
                        },
```

After
```
{
    "characteristicsSorted": {
        "extremelyUnchar": [
            "talkingExpected"
        ],
        "quiteUnchar": [],
        "fairlyUnchar": [],
        "somewhatUnchar": [],
        "neutral": [
            "reassurance"
        ],
        "somewhatChar": [],
        "fairlyChar": [
            "oppositeSex"
        ],
        "quiteChar": [],
        "extremelyChar": []
    }
}
```

## Testing notes

- To avoid schema collisions for other developers, I have created a copy of the ISP experiment on staging that can be controlled separately. Log in to the staging admin interface to get the appropriate experiment ID for your `.env` file.

PR #138, opened 2016-09-16
Feature: De-serialize form data [LEI-194]

samanehsan:
## Purpose
If a user stops the survey mid-way and logs in again to finish, they should see their answers from before.

## Summary of changes
- Include session.expData in `currentFrameConfig` (which gets included with the default params passed to the exp-frame)
- Pass computed properties into the exp-frame
- Define a setter for the `responses` computed property of the `exp-overview` and `exp-rating-form` frames so that the questions are updated with the passed in data.
- Add a flag `loadData` to easily enable/disable loading existing data (Requires https://github.com/CenterForOpenScience/isp/pull/66 to get merged) 

## Note
Holding off on defining a setter for the `exp-card-sort` `responses` computed until https://github.com/CenterForOpenScience/exp-addons/pull/131 is merged

### DIFF
--- a/exp-player/addon/components/exp-card-sort/component.js
+++ b/exp-player/addon/components/exp-card-sort/component.js
@@ -152,14 +152,12 @@ export default ExpFrameBaseComponent.extend({
                   responses['cardSort1'][name] = category.cards.map((cardItem) => cardItem.id);
               }
           }
-          if (this.get('cardSortResponse2')) {
-              cardSortResponse = this.get('cardSortResponse2');
-              // Assumption: this unpacks a list of { categories: {name: name, cards: [cards]} } objects
-              for (let categorySet of cardSortResponse) {
-                  for (let category of categorySet.categories) {
-                      let name = category.name.split('.').pop();
-                      responses['cardSort2'][name] = category.cards.map((cardItem) => cardItem.id);
-                  }
+          cardSortResponse = this.get('buckets2');
+          // Assumption: this unpacks a list of { categories: {name: name, cards: [cards]} } objects
+          for (let categorySet of cardSortResponse) {
+              for (let category of categorySet.categories) {
+                  let name = category.name.split('.').pop();
+                  responses['cardSort2'][name] = category.cards.map((cardItem) => cardItem.id);
               }
           }
           return responses;
@@ -227,7 +225,6 @@ export default ExpFrameBaseComponent.extend({
       window.scrollTo(0,0);
     },
     continue() {
-      this.set('cardSortResponse2', this.buckets2);
       this.send('next');
     }
   },
--- a/exp-player/addon/components/exp-card-sort/component.js
+++ b/exp-player/addon/components/exp-card-sort/component.js
@@ -134,20 +134,36 @@ export default ExpFrameBaseComponent.extend({
   }),
 
   // Represent the sorted cards in a human-readable format for storage in the database
-  responses: Ember.computed('cardSortResponse', function () {
-      let cardSortResponse = this.get('cardSortResponse');
-
-      // Final data should be returned as object {categoryName: [cardIdentifiers] }
-      let responses = {};
-
-      // Assumption: this unpacks a list of { categories: {name: name, cards: [cards]} } objects
-      for (let categorySet of cardSortResponse) {
-          for (let category of categorySet.categories) {
-              let name = category.name.split('.').pop();
-              responses[name] = category.cards.map((cardItem) => cardItem.id);
+  responses: Ember.computed('cardSortResponse', 'cardSortResponse2', {
+      get(key) {
+          // Final data for each card-sort should be returned as object {categoryName: [cardIdentifiers] }
+          let responses = {
+              cardSort1: {},
+              cardSort2: {}
+          };
+          let cardSortResponse;
+          if (this.get('cardSortResponse')) {
+              cardSortResponse = this.get('cardSortResponse');
+              for (let category of cardSortResponse) {
+                  let name = category.name.split('.').pop();
+                  responses['cardSort1'][name] = category.cards.map((cardItem) => cardItem.id);
+              }
           }
+          if (this.get('cardSortResponse2')) {
+              cardSortResponse = this.get('cardSortResponse2');
+              // Assumption: this unpacks a list of { categories: {name: name, cards: [cards]} } objects
+              for (let categorySet of cardSortResponse) {
+                  for (let category of categorySet.categories) {
+                      let name = category.name.split('.').pop();
+                      responses['cardSort2'][name] = category.cards.map((cardItem) => cardItem.id);
+                  }
+              }
+          }
+          return responses;
+      },
+      set(key, value) {
+          return value;
       }
-      return responses;
   }),
 
   isValid: Ember.computed(
@@ -202,12 +218,14 @@ export default ExpFrameBaseComponent.extend({
       target.unshiftObject(card);
     },
     nextPage() {
+      this.set('cardSortResponse', Ember.copy(this.buckets, true));
+      this.send('save');
       this.set('page', 'cardSort2');
       this.sendAction('updateFramePage', 1);
       window.scrollTo(0,0);
     },
     continue() {
-      this.set('cardSortResponse', this.buckets2);
+      this.set('cardSortResponse2', this.buckets2);
       this.send('next');
     }
   },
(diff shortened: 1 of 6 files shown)


exp-player/addon/components/exp-overview/component.js
```diff
@@ -110,13 +110,21 @@ export default ExpFrameBaseComponent.extend(Validations, {
...
     }),
-    responses: Ember.computed(function() {
-      var questions = this.get('questions');
-      var responses = {};
-      for (var i=0; i < questions.length; i++) {
-        responses[i] = questions[i].value;
-      }
-      return responses;
+    responses: Ember.computed({
+      get(key) {
```
samchrisinger: Please fix spacing (4 for indents)

samanehsan: Fixed!

exp-player/addon/components/exp-card-sort/component.js
```diff
@@ -134,20 +134,36 @@ export default ExpFrameBaseComponent.extend({
...
-      let responses = {};
-
-      // Assumption: this unpacks a list of { categories: {name: name, cards: [cards]} } objects
-      for (let categorySet of cardSortResponse) {
-          for (let category of categorySet.categories) {
-              let name = category.name.split('.').pop();
-              responses[name] = category.cards.map((cardItem) => cardItem.id);
+  responses: Ember.computed('cardSortResponse', 'cardSortResponse2', {
+      get(key) {
+          // Final data for each card-sort should be returned as object {categoryName: [cardIdentifiers] }
```
abought: Does this comment need to be updated in any way?

exp-player/addon/components/exp-card-sort/component.js
```diff
@@ -134,20 +134,36 @@ export default ExpFrameBaseComponent.extend({
...
+              for (let categorySet of cardSortResponse) {
+                  for (let category of categorySet.categories) {
+                      let name = category.name.split('.').pop();
+                      responses['cardSort2'][name] = category.cards.map((cardItem) => cardItem.id);
+                  }
+              }
           }
+          return responses;
+      },
+      set(key, value) {
```
abought: Can you please clarify why this setter is necessary?

exp-player/addon/components/exp-card-sort/component.js
```diff
@@ -202,12 +218,14 @@ export default ExpFrameBaseComponent.extend({
       target.unshiftObject(card);
     },
     nextPage() {
+      this.set('cardSortResponse', Ember.copy(this.buckets, true));
```
abought: If `cardSortResponse` is meant to track the data in `buckets`, why not do that as a computed property instead of copying values?

Also, do you need to use `this.get('buckets')`? It looks like this may be setting a value to `undefined`.

exp-player/addon/components/exp-card-sort/component.js
```diff
@@ -202,12 +218,14 @@ export default ExpFrameBaseComponent.extend({
...
+      this.send('save');
       this.set('page', 'cardSort2');
       this.sendAction('updateFramePage', 1);
       window.scrollTo(0,0);
     },
     continue() {
-      this.set('cardSortResponse', this.buckets2);
+      this.set('cardSortResponse2', this.buckets2);
```
abought: See above comment (line 221)

exp-player/addon/components/exp-card-sort/component.js
samanehsan: Overall, if a user exits out of the survey before completing it and then they log back in, they should see all of their previous answers. 

So the setter (which should be a TODO because I'm still writing it!) will need to de-serialize the expData and update the values of the buckets to show the user's previous answers.

But now I'm thinking, maybe it would be easier to take the user directly to the page they were at before exiting?

abought: Interesting question!

I recall you mentioning that ISP is a project that asks users to respond several times in one day. Exactly how long do you expect to go between sessions? (do we want to support completing the form over several visits?)

If we expect a quick return, resuming at the last page makes a lot of sense. If they spend several days between visits, we may just want them to start at the beginning.

exp-player/addon/components/exp-card-sort/component.js
samanehsan: I can't seem to get the `cardSortResponse` computed property to update -- it's watching `'buckets.0.cards.[]', 'buckets.1.cards.[]', 'buckets.2.cards.[]'`

abought: Hmm. It's possible that computed properties work weirdly with such deeply nested items. Long term, we may want to change how buckets are implemented.

One way around this for now: consider making `responses` volatile. Instead of watching a variable, it'll simply be recalculated every time you ask for it.

exp-player/addon/components/exp-card-sort/component.js
samanehsan: Each user is only allowed to complete the ISP survey once, but if they exit before completing the survey they should be able to log back in and finish it. I'm not sure how long people will have to take the study.

exp-player/addon/components/exp-frame-base.js
```diff
@@ -64,19 +64,16 @@ export default Ember.Component.extend({
             this.set('id', `${kind}-${frameIndex}`);
         }
     },
-    setupParams(params, clean) {
+    setupParams(clean) {
         // Add config properties and data to be serialized as instance parameters (overriding with values explicitly passed in)
-        params = params || this.get('frameConfig');
+       var params = this.get('frameConfig');
```
abought: Can you please comment on the reason for removing the argument?

exp-player/addon/components/exp-player.js
```diff
@@ -91,7 +91,16 @@ able to continue the study.
     currentFrameConfig: Ember.computed('frames', 'frameIndex', function() {
         var frames = this.get('frames') || [];
         var frameIndex = this.get('frameIndex');
-        return frames[frameIndex];
+        var config = frames[frameIndex];
```
abought: Rather than bundling the frame data into the configuration, why not pass a second variable into the component? It would also help to minimize side effects- if possible, we'd like to avoid changing the base behavior of frames that are part of other experiments (like Lookit) that share a codebase.

exp-player/addon/components/exp-frame-base.js
```diff
@@ -64,19 +64,16 @@ export default Ember.Component.extend({
...
 
         var defaultParams = {};
         Object.keys(this.get('meta.parameters').properties || {}).forEach((key) => {
             defaultParams[key] = this.get(`meta.parameters.properties.${key}.default`);
         });
 
         Object.keys(this.get('meta.data').properties || {}).forEach((key) => {
-            if (this[key] && this[key].isDescriptor) {
```
abought: Can you please comment on the reason for removing this check?

It appears to have been intentionally added in 7138215014275828dd290e7fde72848a0670c5a0 , with the commit message 
"Dont overwrite computed properties".

exp-player/addon/components/exp-frame-base.js
samanehsan: `setupParms` is only used once [here](https://github.com/CenterForOpenScience/exp-addons/blob/4f1095706ecdae524dc6ef528a4b1c132fb95e4c/exp-player/addon/components/exp-frame-base.js#L54) and `params` is `null` so it doesn't seem necessary. Sam suggested it may be a remnant of old code.

exp-player/addon/components/exp-frame-base.js
samanehsan: This check prevents computed properties from getting passed into the exp-frame. As a result, the value of `responses` in expData will not get passed into the frame, and the user will not see their previous answers. A side-effect of this is that computed properties on exp-frames will need setters.

abought: Hmm. Per previous suggestion, I'm actually not in favor of passing the responses to the frame via this mechanism anyway.

The JSON payload we pass to a frame is meant for configuration, not data- putting the previous session responses into a separate variable passed in to the component might be preferred.

Besides not breaking compatibility with other experiments, this also helps us to avoid bugs that could happen if a field in the server stored data happened to match the name of a variable used for configuration. It's a good idea to keep different data in separate buckets if possible.

samanehsan:
Done making those changes @abought! 🎉

PR #141, opened 2016-09-20
Add calibration videos and confirmation code for continued piloting

kimberscott:
Refs: https://openscience.atlassian.net/browse/LEI-
Companion to:

### DIFF
--- a/exp-player/addon/randomizers/pref-phys-pilot.js
+++ b/exp-player/addon/randomizers/pref-phys-pilot.js
@@ -324,7 +324,7 @@ function toFrames(frameId, eventVideos, BASE_DIR) {
         var allMusic = ['music_01', 'music_02', 'music_03', 'music_04', 'music_06', 'music_07', 'music_09', 'music_10'];
         var musicName = allMusic[Math.floor(Math.random() * allMusic.length)];
 
-        returnFrame = {
+        var returnFrame = {
             kind: 'exp-video-physics',
             id: `${frameId}`,
             autoplay: true,
@@ -351,7 +351,7 @@ function toFrames(frameId, eventVideos, BASE_DIR) {
         };
 
         // FOR PILOT ONLY: replace fall videos with calibration
-        if (e.compType == 'fall') {
+        if (e.compType === 'fall') {
             returnFrame.sources = videoSourceObjs(
                 BASE_DIR + 'stimuli/attention/',
                 'calibration');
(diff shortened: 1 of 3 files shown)


exp-player/addon/randomizers/pref-phys-pilot.js
```diff
@@ -344,12 +343,25 @@ function toFrames(frameId, eventVideos, BASE_DIR) {
...
                 BASE_DIR + 'stimuli/' + features.eventType + '/',
                 e.altName, true)
         };
+
+        // FOR PILOT ONLY: replace fall videos with calibration
+        if (e.compType === 'fall') {
+            returnFrame.sources = videoSourceObjs(
+                BASE_DIR + 'stimuli/attention/',
+                'calibration');
+            returnFrame.sources = videoSourceObjs(
```
abought: Is this repetition intentional? (lines 355-357 ; 358-360)

kimberscott: Ack! No, it is not. 358 should define
`
returnFrame.altSources = [...]`

abought: If you're not near a computer, I'll go ahead and make those changes myself...?

Before we go to production on this, I suggest trying it out on the staging server tonight (along with the sentry logging feature). That will avoid surprises when this goes to production.

Naive question, but I assume that `e.compType === 'fall'` is something that only happens in the pilot? (I ran through a sample study earlier on staging, and saw some items falling. I may be overinterpreting that)

kimberscott: I've fixed it & can try it on the staging server tonight! (I did try the changes out locally, but forgot to try pausing one of the calibration videos. Thanks so much for the careful review!)

`e.compType === 'fall'` can happen on both the pilot and regular versions of the study, it's just that here we'll substitute in some calibration videos in place of the objects falling. For the actual study we care about kids' preferential looking during those videos; for the pilot, we just care about whether they find the videos overall reasonably engaging and whether we can code their direction of gaze. (The calibration videos are showing up correctly using this version of the prefphyspilot randomizer - every 4th video is a calibration video.)

abought:
I'll go ahead and merge this. I've put in a request to the person in charge of the server to update something that it will need to work, and I'll send you a note when that's ready.

Incidentally, did you and Sam work out a faster way to run through an experiment? It's something of a slow process at the moment. In particular, my completed sessions weren't showing up in `staging-experimenter`, so I want to be 100% sure that isn't a recurring problem.

kimberscott:
Okay, thanks!

To run through the experiment faster, we'd previously been setting MAX_VIDEOS to 4 or so (line 369 of pref-phys-pilot.js) and testLength to 5 s (line 331). It would be much cleaner to make those into properties that could be passed in by the json schema defining the experiment, but I wasn't sure how. 

I've previously had trouble with sessions not showing up in experimenter, but don't actually use that feature--I've just been running updates using the code in lookit/scripts in my coding-workflow branch of lookit. I do see sessions showing up right away in the consent spreadsheets we generate.

PR #143, opened 2016-09-22
Feature: fix continuing study [LEI-210]

samanehsan:
Refs: https://openscience.atlassian.net/browse/LEI-210
Companion to: https://github.com/CenterForOpenScience/isp/pull/68

## Purpose
When a user continues their session after exiting, they should be redirected to the last frame/page they were on.

## Summary of changes
- Save `frameIndex` and `framePage` on the session
- Make card-sort frame use `framePage` instead of defining it's own `page` property
- Simplify loading card-sort data
- Only load `exp-rating-form` data once instead of on each `framePage`


exp-models/addon/models/session.js
```diff
@@ -18,6 +18,9 @@ export default DS.Model.extend(AnonJamModel, {
     experimentId: DS.attr('string'),
     experimentVersion: DS.attr('string'),  // TODO: Currently this field is not acted on in any way
 
+    frameIndex: DS.attr(),
```
abought: These new model attributes appear to be ISP-specific fields. Can you elaborate on why these properties are in the shared model in `exp-addons`, rather than in a customized subclass of this model in the ISP repo?

(You can extend the base model using the same technique you used for the authenticators early this week- import and extend)

abought: Also, at minimum, a source code comment explaining the purpose of these parameters would be helpful, for anyone trying to understand what these fields are later.

(given how we approach pagination in this app, the meaning of `frameIndex` is especially murky. Page number? Number of pages? Number of the subframe within the frame within the page pretending to be 10 frames?)

exp-player/addon/components/exp-player.js
```diff
@@ -148,6 +148,9 @@ able to continue the study.
                 this._transition();
                 this.set('frameIndex', frameIndex + 1);
                 this.set('framePage', 0);
+                this.get('session').set('frameIndex', frameIndex + 1);
+                this.get('session').set('surveyPage', 0);
+                this.get('session').save();
```
abought: I think session may already be saved somewhere else, whenever we advance a frame. Is this causing data to be saved twice for each click of "next"?

exp-player/addon/components/exp-player.js
```diff
@@ -148,6 +148,9 @@ able to continue the study.
                 this._transition();
                 this.set('frameIndex', frameIndex + 1);
                 this.set('framePage', 0);
+                this.get('session').set('frameIndex', frameIndex + 1);
```
abought: What is the advantage of tracking the same property twice? For example, are there specific design barriers to directly binding the value from the session to the controller? (`Ember.computed.alias` or `Ember.computed.or`, etc)

Two way bindings can be weird sometimes, but this implementation is pretty explicit that whenever the internal state variable changes, the session is also modified and saved. In that case, a two way data binding might be ok and not a major design change.

exp-player/addon/components/exp-player.js
abought: In particular, remember that this is meant to be a somewhat general infrastructure. Not every experiment on this platform may want to save data as the user goes along- some may want to only save at end. 

We should be careful about modifying the base `exp-player` too much to suit the use case of one particular experiment. (And I say that for my work on Lookit as well... we can learn from past experience here)

exp-models/addon/models/session.js
abought: Right- it was tracked internally, but that's a separate matter from deciding that every session for any experiment needs to track and save this data... 

Main areas of potential confusion include:
1. Many experiments are not surveys
2. If frameIndex is the last "page", then what is "survey page"?

exp-player/addon/components/exp-player.js
samanehsan: I am getting the value of `frameIndex` from the session but `session.frameIndex` does not automatically get updated when the value of the `frameIndex` changes: https://github.com/CenterForOpenScience/isp/pull/68/files#diff-c5ef69c69ee1fddc9d65f6d3878873dfR8

PR #145, opened 2016-09-23
Rating Form UI changes [LEI-213]

samanehsan:
Refs: https://openscience.atlassian.net/browse/LEI-213

## Changes
- Increase spacing of radio buttons in questions 2, 4, & 6
- Fix scales in questions 2, 6 & 9

### DIFF
--- a/exp-player/addon/components/exp-rating-form/component.js
+++ b/exp-player/addon/components/exp-rating-form/component.js
@@ -264,19 +264,19 @@ var questions = [
             formatLabel: 'format-label',
             labels: [
                 {
-                    rating: 0,
+                    rating: 'number1',
                     label: 'measures.questions.2.options.never'
                 },
                 {
-                    rating: 2,
+                    rating: 'number3',
                     label: 'measures.questions.2.options.hardlyEver'
                 },
                 {
-                    rating: 4,
+                    rating: 'number5',
                     label: 'measures.questions.2.options.occasionally'
                 },
                 {
-                    rating: 6,
+                    rating: 'number7',
                     label: 'measures.questions.2.options.quiteOften'
                 }
             ]
@@ -314,11 +314,11 @@ var questions = [
             formatLabel: 'measure-four',
             labels: [
                 {
-                    rating: 0,
+                    rating: 'number0',
                     label: 'measures.questions.4.options.unwilling'
                 },
                 {
-                    rating: 10,
+                    rating: 'number10',
                     label: 'measures.questions.4.options.fullyPrepared'
                 }
             ]
@@ -351,11 +351,11 @@ var questions = [
                 formatLabel: 'measure-six',
                 labels: [
                     {
-                        rating: 0,
+                        rating: 'number1',
                         label: 'measures.questions.6.items.1.options.notHappy'
                     },
                     {
-                        rating: 6,
+                        rating: 'number7',
                         label: 'measures.questions.6.items.1.options.veryHappy'
                     }]
             },
@@ -367,11 +367,11 @@ var questions = [
                 formatLabel: 'measure-six',
                 labels: [
                     {
-                        rating: 0,
+                        rating: 'number1',
                         label: 'measures.questions.6.items.2.options.lessHappy'
                     },
                     {
-                        rating: 6,
+                        rating: 'number7',
                         label: 'measures.questions.6.items.2.options.moreHappy'
                     }]
             },
@@ -383,11 +383,11 @@ var questions = [
                 formatLabel: 'measure-six',
                 labels: [
                     {
-                        rating: 0,
+                        rating: 'number1',
                         label: 'measures.questions.6.items.4.options.notAtAll'
                     },
                     {
-                        rating: 6,
+                        rating: 'number7',
                         label: 'measures.questions.6.items.4.options.aGreatDeal'
                     }]
             },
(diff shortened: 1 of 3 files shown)


exp-player/addon/components/exp-rating-form/component.js
```diff
@@ -378,13 +380,14 @@ var questions = [
                 scale: SEVEN_POINT_SCALE,
                 value: null,
                 labelTop: false,
+                formatLabel: 'measure-six',
                 labels: [
                     {
                         rating: 0,
```
abought: We want the labels to say 1-7... Should the internal values we map to that scale match? (setting a range 0-6 seems a bit odd)

exp-player/addon/components/exp-rating-form/component.js
```diff
@@ -192,8 +192,8 @@ const TEN_POINT_SCALE = [
     'number9',
     'number10'
 ];
-const SEVEN_POINT_SCALE = TEN_POINT_SCALE.slice(0, 8);
-const NINE_POINT_SCALE = TEN_POINT_SCALE.slice(0, 10);
+const SEVEN_POINT_SCALE = TEN_POINT_SCALE.slice(1, 8);
```
abought: These scales are different lengths before (0,8) and after (1,8). How did this work out to the same number of items??

exp-player/addon/styles/components/isp-radio-group.scss
```diff
@@ -24,6 +24,11 @@
   margin-top: -15px;
 }
 
-.label-spacing {
-  max-width: 6%;
+.measure-four {
+  width: 6%;
```
abought: Can you please clarify how you arrived at this number?  These styles seem to be very question-specific. If this is to label something reused repeatedly (eg 7 point scale), you may not want the style to be tied to just one question on a form.

exp-player/addon/components/exp-rating-form/component.js
samanehsan: I accidentally included one extra item (instead of 0-7, the scale is supposed to go from 1-7)

exp-player/addon/styles/components/isp-radio-group.scss
samanehsan: I tried re-using classes when possible (e.g. The `format-label` class is used for questions 2 & 3) but the formatting looked off in some cases

exp-player/addon/components/exp-rating-form/component.js
samanehsan: I was mapping the labels to the correct radio button option based on index - it looks odd since some of the scales start at 1 and not 0. I could make this more explicit by comparing the value instead of the index.

samanehsan:
Done responding to your comments/questions @abought! Let me know if I should change the rating labels to map to option values instead of indices 🍂

abought:
> Done responding to your comments/questions @abought! Let me know if I should change the rating labels to map to option values instead of indices 🍂

I'll leave this slightly to your discretion- but maintaining two distinct sets of numbers associated with each item is slightly confusing, and seems like it could make the code harder to maintain. Somewhere at the back of my mind, this seems to hint that the code is duplicating more than necessary.

Let me know what you decide.

exp-player/addon/styles/components/isp-radio-group.scss
abought: If two sets of scales had the same number of options, why did the formatting "look off" for one? Is there an example?

Without specific examples to see, just as general notes: what I've found in the past is that if two things were otherwise identical, then if you had to "special case" the display of one of them (with extra styles), it meant there was some basic underlying difference responsible for the display difference. The more specific trial-and-error display tweaks you have to make, the more likely it is that your page display is locked to a set of assumptions that could break for someone else. (smaller monitor, larger text, etc)

samanehsan: In this case, I added the specific class for measure four because it has 11 options, and needed a smaller width than the existing css classes. 

I added the `measure-six` class to format this question:  
screen shot 2016-09-26 at 4 18 58 pm

If I use the existing `format-label` class it will look like this: 
screen shot 2016-09-26 at 4 22 23 pm

Personally, I like the top one better because it's more spaced out, which is why there is a custom css class.

samanehsan:
@abought, done responding! I changed the option labels to map to the option values to make that clearer. ☔

exp-player/addon/components/exp-rating-form/component.js
```diff
@@ -264,19 +264,19 @@ var questions = [
             formatLabel: 'format-label',
             labels: [
                 {
-                    rating: 0,
+                    rating: 'number1',
```
abought: I'll merge this as is for now, but we should really start considering how data is serialized. (Are they going to want to do analysis on "number0" in the CSV file? Or are they going to want a computed program that adds up numeric integer values for each field?)

samanehsan: I think the numbers should get serialized as the actual number (e.g. 1 instead of "number1").

exp-player/addon/styles/components/isp-radio-group.scss
abought: To be honest I'm not totally seeing the difference in spacing in the two screenshots.

I'll merge this because we need to go ahead for UCR review, but if there are 11 equally spaced options, then I'm not sure how we arrived at 6% for the final width. (100% / 11 = ~9%)...

PR #147, opened 2016-09-27
Fix card-sort serialization [LEI-216]

samanehsan:
Refs: https://openscience.atlassian.net/browse/LEI-216

## Purpose
Instead of serializing card-sort data into the format: `{bucket0: [card1, card2, card3...]}`
change it to `{card1: 0, card2: 0, card3: 0, ...}`

## Changes
- Change `meta.data.properties` & `responses` computed

### DIFF
--- a/exp-player/addon/components/exp-card-sort/component.js
+++ b/exp-player/addon/components/exp-card-sort/component.js
@@ -136,16 +136,17 @@ export default ExpFrameBaseComponent.extend({
   // Represent the sorted cards in a human-readable format for storage in the database
   responses: Ember.computed(function() {
       // Final data should be returned as {
-      //    cardSort1: object {categoryName: [cardIdentifiers] }
-      //    cardSort2: object {categoryName: [cardIdentifiers] }
+      //    cardSort1: object {cardId: categoryId,...}
+      //    cardSort2: object {cardId: categoryId,...}
       // }
       let responses = {};
       let cardSortResponse = this.get('cardSortResponse');
       if (cardSortResponse) {
           responses['cardSort1'] = {};
-          for (let category of cardSortResponse) {
-              let name = category.name.split('.').pop();
-              responses['cardSort1'][name] = category.cards.map((cardItem) => cardItem.id);
+          for (var i=0; i < cardSortResponse.length; i++) {
+              for (let card of cardSortResponse[i].cards) {
+                  responses['cardSort1'][card.id] = i + 1;
+              }
           }
       }
       if (this.get('page') === 'cardSort2') {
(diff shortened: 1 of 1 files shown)


exp-player/addon/components/exp-card-sort/component.js
```diff
@@ -315,46 +326,8 @@ export default ExpFrameBaseComponent.extend({
...
-                          type: 'array'
-                      },
-                      quiteChar: {
-                          type: 'array'
-                      },
-                      extremelyChar: {
-                          type: 'array'
-                      }
+                  responses: {
+                      type: 'object'
```
abought: Is the server in any way responsible for validating that all questions are filled out? If we totally control the front-end code, and how users interact with the survey, this may be acceptable- just so long as the decision to remove backend validation is a conscious and active design choice.

exp-player/addon/components/exp-card-sort/component.js
```diff
@@ -136,26 +136,28 @@ export default ExpFrameBaseComponent.extend({
...
       }
       if (this.get('page') === 'cardSort2') {
           cardSortResponse = this.get('buckets2');
           responses['cardSort2'] = {};
           // Assumption: this unpacks a list of { categories: {name: name, cards: [cards]} } objects
           for (let categorySet of cardSortResponse) {
-              for (let category of categorySet.categories) {
-                  let name = category.name.split('.').pop();
-                  responses['cardSort2'][name] = category.cards.map((cardItem) => cardItem.id);
+              for (var j=0; j < categorySet.categories.length; j++) {
```
abought: Does this extra level of iteration indicate a change in the data structure we're operating on? Can you please verify that the source code comments are accurate and up-to-date?

Is the final output of this computed a category ID (as suggested in line 139), or a numeric value corresponding to the category?

samanehsan: Yes, `buckets2` is now the same data structure as defined in `meta.properties.buckets2`.

abought: Can you please update the source code comments in ways that will make this code more maintainable? (or at least to not be inaccurate)

Even if not a full YUIDoc comment, future developers who work with this code should not have to reverse-engineer any mystery data structures to understand what a block of code does / returns.

abought:
> The validation is only on the front-end right now, but it is probably better to have both so I'll add validation to thecontinue actions

Server side validation would not happen in an emberJS action... and for 90-180 questions I agree that maintaining it in the front-end would be fairly painful. 

We can discuss how server side validation works later to see if it's right for this use case. (it's possible it may not be- but again, that should be a conscious choice, not something we remove and forget about).

samanehsan:
Oh, I see what you're saying. There would have to be 90 properties listed for both cardSort1 and cardSort2 in `meta.data.properties`

exp-player/addon/components/exp-card-sort/component.js
samanehsan: Done!

PR #149, opened 2016-09-28
Use integers for options [LEI-217]

samanehsan:
Refs: https://openscience.atlassian.net/browse/LEI-217
Companion to: https://github.com/CenterForOpenScience/isp/pull/72

## Purpose
Since the translation strings for numbers will be just the number (e.g. "0" instead of
"number0"), use integers in the options scales so that those values are serialized as integers.

## Testing notes
Requires changes in https://github.com/CenterForOpenScience/isp/pull/72

### DIFF
--- a/exp-player/addon/components/exp-overview/component.js
+++ b/exp-player/addon/components/exp-overview/component.js
@@ -114,7 +114,12 @@ export default ExpFrameBaseComponent.extend(Validations, {
         var questions = this.get('questions');
         var responses = {};
         for (var i=0; i < questions.length; i++) {
-            responses[i] = questions[i].value;
+            if (i === 0) {
+                // Convert value to int bc select-input returns a string (e.g. "16" --> 16)
+                responses[i] = parseInt(questions[i].value);
+            } else {
+                responses[i] = questions[i].value;
+            }
         }
         return responses;
     }).volatile(),
--- a/exp-player/addon/components/exp-overview/component.js
+++ b/exp-player/addon/components/exp-overview/component.js
@@ -7,8 +7,7 @@ import {validator, buildValidations} from 'ember-cp-validations';
 function range(start, stop) {
   var options = [];
   for (var i=start; i <= stop; i++) {
-    var key = 'number' + i;
-    options.push(key);
+    options.push(i);
   }
   return options;
 }
@@ -158,7 +157,7 @@ export default ExpFrameBaseComponent.extend(Validations, {
                                type: 'string'
                             },
                             '7': { // how religious?
-                                type: 'string'
+                                type: 'integer'
                             },
                             '8': { // follows religion?
                                 type: 'string'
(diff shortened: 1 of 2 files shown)


samanehsan:
The answer to the [first question](https://github.com/samanehsan/exp-addons/blob/bf8eca8d82c293ef36d41da8700b349aa9edc8a1/exp-player/addon/components/exp-overview/component.js#L138) in the exp-overview form is still getting serialized as a string. @abought, do you think that has something to do with the select drop down?

abought:
> The answer to the first question in the exp-overview form is still getting serialized as a string. @abought, do you think that has something to do with the select drop down?

Well... value you're giving it is a string, right? JSON does have a way to distinguish between stringers and numbers, so I imagine that the values we give it would come through in how we serialize it.

It's hard to tell whether this is a problem with one field, insofar as many of the other fields that use this scale don't check the datatype at all. This might be a problem with the field, or it might just be that this field is the first time we ever looked that closely, and they might all be like this... it's worth digging in a bit if you think the serialization will be a problem.

samanehsan:
@abought, the options that I'm giving to the `select-input` component are integers, but the value that gets saved is a string. When I pass integers as the radio button options, the value I get is an integer. So, I'm not sure what exactly is causing this difference.

samanehsan:
@abought, I did some more investigating and I think I'll need to convert the value of the first question in exp-overview to an integer. Although, it looks like jamdb is not even checking to make sure it is an integer.

abought:
Good to know- and I'm a bit surprised that the type of the value is changing, but converting it wouldn't be too onerous.

As for validation: Jam doesn't check until you tell it to- it doesn't know about source code changes in the ember app magically.  

To trigger server-side validation on this field, you would need to run experimenter locally, and re-save the experiment definition from a computer where the validation was defined in the code for the specified experiment frame.

samanehsan:
@abought, good point - I don't think I re-saved the experiment definition

samanehsan:
Made that change @abought! 🌂

exp-player/addon/components/exp-overview/component.js
```diff
@@ -114,7 +114,12 @@ export default ExpFrameBaseComponent.extend(Validations, {
         var questions = this.get('questions');
         var responses = {};
         for (var i=0; i < questions.length; i++) {
-            responses[i] = questions[i].value;
+            if (i === 0) {
+                // Convert value to int bc select-input returns a string (e.g. "16" --> 16)
+                responses[i] = parseInt(questions[i].value);
```
abought: That's a bit gnarly, but we can improve it in the future.

As I recall, this field has a default value (null): https://github.com/CenterForOpenScience/exp-addons/blob/a3d18733534d6a35db449f276ec3294948037e37/exp-player/addon/components/select-input/template.hbs#L2-L2

Are we comfortable with the default serialization (`NaN`) that would result? It doesn't throw an error, but there is a slight assumption that this is always an integer.

abought: Given the time and date, I'll merge as is for now, but think back on this, and it can be changed in a followup PR if needed.

PR #155, opened 2016-09-30
Feature/validation flag [LEI-231]

samanehsan:
Refs: https://openscience.atlassian.net/browse/LEI-231
Companion to: https://github.com/CenterForOpenScience/isp/pull/76

## Purpose
Add the ability to turn validation off for testing.

## Summary of changes
If the `validate` flag is set to false, allow users to continue through the survey without having to answer every question.

### DIFF
--- a/exp-player/addon/components/exp-card-sort/component.js
+++ b/exp-player/addon/components/exp-card-sort/component.js
@@ -1,6 +1,7 @@
 import Ember from 'ember';
 import ExpFrameBaseComponent from 'exp-player/components/exp-frame-base';
 import layout from './template';
+import config from 'ember-get-config';
 
 
 var cards = {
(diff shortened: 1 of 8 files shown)


exp-player/addon/components/exp-card-sort/component.js
```diff
@@ -168,6 +169,12 @@ export default ExpFrameBaseComponent.extend({
         }
         return responses;
     }).volatile(),
+
+    allowNext: Ember.computed('cards.[]', function() {
+        console.log(this.get('cards').length);
```
abought: Please remove `console.log` statements before committing

Also, I notice the indentation is different between these two lines?

exp-player/addon/components/exp-card-sort/component.js
```diff
@@ -168,6 +169,12 @@ export default ExpFrameBaseComponent.extend({
...
+      return this.get('cards').length === 0 || !config.validate;
```
abought: Should this depend on two conditions both being met? Using "or", it seems that there is a situation where only the first condition would need to be satisfied for this to return true.

(In other words, the `config.validate` flag might be ignored)

samanehsan: This should return true if there are no cards left to sort, or if validation is turned off. I changed it to explicitly check `config.validate` for clarity.

samanehsan:
Done responding @abought! ☔

PR #157, opened 2016-10-04
Fix rating form validation [LEI-232]

samanehsan:
Refs: https://openscience.atlassian.net/browse/LEI-232

## Purpose
Currently, validation is only working on the first `framePage` of the `exp-rating-form`. All validators on the other pages are disabled, but do not get re-enabled when the framePage changes.

## Summary of changes
Add validators for each page that depend on validators for questions on that page. Make allowNext track the page validators and framePage, and return whether the current page's validator is valid.

## Testing notes
The `continue` button should be disabled until all of the questions on the page are answered. However, if the `config.validate` flag is `false`, all of the continue buttons should be enabled.

### DIFF
--- a/exp-player/addon/components/exp-rating-form/component.js
+++ b/exp-player/addon/components/exp-rating-form/component.js
@@ -187,21 +187,27 @@ const NINE_POINT_SCALE = TEN_POINT_SCALE.slice(1, 10);
 
 var generateValidators = function (questions) {
     var validators = {};
+    var pages = {};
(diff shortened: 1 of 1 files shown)


abought:
This is cleaner than we were hoping- nice!

I need to work on a few other tickets, and I think the key to this PR is going to be using it a bunch in different circumstances. I'll go ahead and merge it on the premise that having it in circulation will ensure the best functional review we can offer here- if there are regressions we'll ticket that as a followup.

PR #159, opened 2016-10-04
Add computed for disabling login button [LEI-238]

samanehsan:
Refs: https://openscience.atlassian.net/browse/LEI-238
Companion to: https://github.com/CenterForOpenScience/isp/pull/77

## Purpose
Button is now also disabled if the invalidAuth modal is open.

## Summary of changes
- Add `disabled` computed to replace logic in the template
- Button is now also disabled if the `invalidAuth` modal is open (prevents flashing disabled --> enabled). When the user dismisses the modal, the button will be enabled and they will be able to re-submit the form with a correct username & password.

### DIFF
--- a/exp-models/addon/components/jam-auth/component.js
+++ b/exp-models/addon/components/jam-auth/component.js
@@ -13,7 +13,7 @@ export default Ember.Component.extend({
     authenticating: false,
     invalidAuth: false,
 
-    disabled: Ember.computed('username', 'password', 'authenticating', 'invalidAuth', function() {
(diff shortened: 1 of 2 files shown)


exp-models/addon/components/jam-auth/component.js
```diff
@@ -11,6 +11,11 @@ export default Ember.Component.extend({
     password: null,
 
     authenticating: false,
+    invalidAuth: false,
+
+    allowLogin: Ember.computed('username', 'password', 'authenticating', 'invalidAuth', function() {
```
abought: In the template usage, we disable the login button if this is true.

...So really, should this be named `disableLogin`?

PR #161, opened 2016-10-05
Split birth city & country into two questions [LEI-244]

samanehsan:
Refs: https://openscience.atlassian.net/browse/LEI-244

- Make birth city & country two questions
- Update showOptional computed to track the optional question that is now at questions[9]


abought:
This one really doesn't need much review; in it goes!

PR #162, opened 2016-10-05
Feature/update thank you page [LEI-240]

samanehsan:
Refs: https://openscience.atlassian.net/browse/LEI-240

## Purpose
Update thank-you page: 

## Summary of changes
- Center and enlarge buttons
- Add a picture  
screen shot 2016-10-05 at 5 14 01 pm


exp-player/addon/styles/addon.scss
```diff
@@ -37,6 +37,18 @@
     margin: 5px;
 }
 
+.center {
```
abought: I don't think this PR needs any especially fancy review. My main suggestion is some general advice:

1. Consider whether we want these special-purpose CSS classes- especially with such generic names that might overlap with future additions.

2. Can you accomplish your styling/layout goals using existing bootstrap functionality? If not, is there a better way to keep your CSS isolated? If we eventually refactor the ISP frames into their own repo, how will we make sure that we find all the CSS styles those frames need to work?

Picking apart one big monolith will be really hard, but there's some room to plan ahead.

samanehsan: I can use `text-center` here instead so I removed this class

samanehsan:
Done making those changes @abought! 🍂

exp-player/addon/styles/addon.scss
```diff
@@ -37,6 +37,14 @@
     margin: 5px;
 }
 
+.margin-top-sm {
+    margin-top: 25px;
```
abought: Are these definitions of small and large unique to your use case, or are they more generic? How did you choose these numbers? 

Remember, every single exp-frame will see this style name, and when you move your exp-frames to a different repo*, you might want a way to identify which styles need to go with it. Generic names for special-purpose things will make those styles harder to find later.

Aside from general advice, I'm going to leave the final decision on merging to you on this one, since these are pure UI changes specific to ISP. 



* There's no guarantee we'll be moving exp-frames out of exp-addons and into ISP. But it is a change that we are strongly considering as part of the Lookit platform/ phase 2 discussions, and planning ahead a bit will make our lives easier later.

samanehsan: It's specific to that ISP frame -- so I'll move it into a separate CSS file

samanehsan:
Made that change @abought! 🎃

abought:
These changes should help future refactoring; thanks.

Ultimately the decision to merge this one was yours, so since you indicated it's ready, in we go!

PR #163, opened 2016-10-06
Feature/add rating question [LEI-241]

samanehsan:
Refs: https://openscience.atlassian.net/browse/LEI-241

## Purpose
Add in another optional question under question 10 on the rating-form, below "What aspect are you trying to change?"

## Summary of changes
- Add rating form question that only shows up when a user answers "yes" to question 10:  
screen shot 2016-10-06 at 11 07 33 am

- Adjust spacing on question four so that there is less space between the question and answer options:  
screen shot 2016-10-06 at 11 07 19 am

## Notes
- I noticed the question's wording seems incorrect so I asked for clarification (but that change would occur in `translations.js`

### DIFF
--- a/exp-player/addon/components/exp-rating-form/component.js
+++ b/exp-player/addon/components/exp-rating-form/component.js
@@ -301,7 +301,7 @@ var questions = [
         scale: TEN_POINT_SCALE,
         options: {
             labelTop: false,
-            formatLabel: 'measure-four',
+            formatLabel: 'measure-four negative-margin-top',
             labels: [
                 {
                     rating: 0,
--- a/exp-player/addon/components/exp-rating-form/component.js
+++ b/exp-player/addon/components/exp-rating-form/component.js
@@ -464,14 +464,45 @@ var questions = [
         scale: ['measures.questions.11.options.yes', 'measures.questions.11.options.no'],
         items: [{
             type: 'radio',
-            value: null
+            value: null,
+            formatLabel: 'negative-margin-top'
         },
-            {
-                type: 'input',
-                description: 'measures.questions.12.label',
-                value: null,
-                optional: true
-            }]
+        {
+            type: 'textarea',
+            description: 'measures.questions.12.label',
+            value: null,
+            optional: true
+        },
+        {
+            type: 'radio',
+            description: 'measures.questions.18.label',
+            value: null,
+            optional: true,
+            labelTop: false,
+            scale: NINE_POINT_SCALE,
+            labels: [
+                {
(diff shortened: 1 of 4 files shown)


abought:
I'm slightly concerned about the one-off custom styling here. Why do we only need to style only this specific question? 

There are many places in ISP where we put a row of selection buttons under a sentence- what is unique about these specific questions to merit the visual special-casing? Why aren't all the other radio button groups restyled?

What happens when the browser is resized, or font size changed? If the difference is "how long the label text is", remember that we will be translating it into other languages- we should not special case something just because it looks that way in english.

samanehsan:
@abought, the difference is in the format of the questions. For example, question 10 does not have multiple sub-items beneath it like question 11 does. Since there is no sentence to break up that space in between the question and radio buttons, it looks awkward without the styling:   
screen shot 2016-10-07 at 2 26 12 pm

abought:
> Since there is no sentence to break up that space in between the question and radio buttons, it looks awkward without the styling:

Since the radio button group component handles the drawing of labels, perhaps it would make more sense (and less code) to handle the case of no labels in the component, rather than a special case for every question that works this way?

samanehsan:
The radio button group component only handles drawing the radio button labels (not the item labels) those get listed in the `exp-rating-form` component: https://github.com/samanehsan/exp-addons/blob/70fb0ca3477fe61fde8c94fca1ca4c87cbd63567/exp-player/addon/components/exp-rating-form/template.hbs#L9

abought:
> The radio button group component only handles drawing the radio button labels (not the item labels) those get listed in the exp-rating-form component: https://github.com/samanehsan/exp-addons/blob/70fb0ca3477fe61fde8c94fca1ca4c87cbd63567/exp-player/addon/components/exp-rating-form/template.hbs#L9

It seems a bit weird to me that the logic for drawing/labeling these buttons is spread out between the rating form and the radio button widgets. We *can* special case each individual question that needs to be drawn this way, but it adds code, and is a bit more brittle than if a component was fully able to do the things that are its responsibility (eg labels). In particular, specifying a particular pixel size for the margins seems like it could break if you resize or change the length of the text.

If this is the solution you need for now based on client schedule and your own queue, then go ahead and merge accordingly. But we should be cognizant that this isn't a long-term solution if we want to re-use this component later.

samanehsan:
I'll go ahead and merge this now, and send in a follow-up PR with a refactor. Since the amount of space below the question depends on whether it has multiple items, I can move the item labels into the radio-group component and do the check there.

PR #166, opened 2016-10-11
Export jam-jwt authenticator [LEI-224]

samanehsan:
Refs: https://openscience.atlassian.net/browse/LEI-224
Companion to: https://github.com/CenterForOpenScience/isp/pull/83

## Purpose
Export jam-jwt authenticator, which gets extended by ISP's corresponding authenticator: https://github.com/CenterForOpenScience/isp/pull/83/files#diff-cb23210aeda9c128dc6c539f7e0c22acR4


samanehsan:
This change is not necessary to import the `jam-jwt`, as I previously thought

PR #167, opened 2016-10-12
Confirmation dialog when user withdraws their data

abought:
Refs: https://openscience.atlassian.net/browse/LEI-249
Companion to: TBD (will need to update Lookit submodule commit)

## Purpose
Improve the experiment exit survey.

## Summary of changes
- [x] Add confirmation dialog if user wants to withdraw their results from an experiment
- [x] Make sure that sessions are correctly marked as completed at conclusion of the study
- ~~Fix "Number sessions completed" indicator to filter by completed flag (moved to separate ticket)~~

screen shot 2016-10-12 at 10 54 21 pm

## Testing notes
This affects the last frame (exit survey) when completing a study.

To test locally, you can change code to make the study run faster. See JIRA ticket for notes on what variables to modify.

## Intended behaviors
- [x] When user selects "withdrawal", they are given a confirmation box.
  - [x] If they click "Cancel", they return to section 1 of the exit survey and can change their mind
  - [x] If they click "withdraw", they move to section 2
  - [x] If they click cancel, then continue without changing the checkbox setting, they will see the modal again (modal appears every time they try to advance with checkbox active)
- [x] When "withdrawal" box is not checked, the user sees no modal.

### DIFF
--- a/exp-player/addon/components/exp-exit-survey-pilot.js
+++ b/exp-player/addon/components/exp-exit-survey-pilot.js
@@ -81,12 +81,22 @@ export default ExpFrameBaseComponent.extend(Validations, FullScreen, {
     },
     today: new Date(),
     section1: true,
+    showWithdrawalConfirmation: false,
     showValidation: false,
     actions: {
+        advanceToProgressBar() {
+            // Move from section 1 (survey) to section 2 (progress bar/ finish button)
+            this.set('section1', false);
+            this.send('save');
+        },
         continue () {
+            // Check whether exit survey is valid, and if so, advance to next screen
             if (this.get('validations.isValid')) {
-                this.set('section1', false);
-                this.send('save');
+                if (this.get('withdrawal')) {
+                    this.set('showWithdrawalConfirmation', true);
+                } else {
+                    this.send('advanceToProgressBar');
+                }
             } else {
                 this.set('showValidation', true);
             }
--- a/exp-player/addon/components/exp-exit-survey.js
+++ b/exp-player/addon/components/exp-exit-survey.js
@@ -81,12 +81,22 @@ export default ExpFrameBaseComponent.extend(Validations, FullScreen, {
     },
     today: new Date(),
     section1: true,
+    showWithdrawalConfirmation: false,
     showValidation: false,
     actions: {
+        advanceToProgressBar() {
+            // Move from section 1 (survey) to section 2 (progress bar/ finish button)
+            this.set('section1', false);
+            this.send('save');
+        },
         continue () {
+            // Check whether exit survey is valid, and if so, advance to next screen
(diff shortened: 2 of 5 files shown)


abought:
Will merge this pending MIT approval of language in modal- we'd like to be sure that the instructions and button labels are clear for the intended audience. ( @kimberscott )

Changes to the progress bar screen will happen in a separate ticket.

It is worth noting that this PR as written only changes one component. If we are maintaining multiple separate versions of the exit survey (eg for piloting), we may wish to consider how to DRY this code for maintainability.

kimberscott:
Language: can we change this slightly to "You have chosen to withdraw your video data from the study. Any video beyond your consent recording will be deleted without viewing. Are you sure?"

Components: Can this change also be applied to exp-exit-survey-pilot, so we're using it during the next bit of piloting? (To avoid repetition we could generalize the exit-survey frame with an option for whether to show progress message & bar at all, and just put that in an {{#if showProgress}} block. But it's also only two frames.)

Behavior: Does confirmation popup appear when they click "submit" with the "withdrawal" box checked, or immediately upon checking the box? Could it be the former so they can submit feedback either way? (The "withdraw" option is for video only; we actually don't offer a way to remove ALL data from the study, and may want to read what they say about e.g. why they left early.)

abought:
## Language
Great point! I'll make that change now.

## Behavior
Currently, this pops up only when they click "Return to Lookit" (the button that submits the survey).

I will change that to happen immediately when they click the checkbox instead.

## Components
Per duplicate frames, it's mainly an issue in the sense that code will diverge over time, which makes it increasingly challenging to backport every single bugfix to multiple files. Programmers look to avoid duplicate code if possible because it creates a place for bugs and regressions to hide. (we don't like giving you buggy code)

It sounds like we have feature priorities first, but I'll ticket a refactor of the survey code to support a configurable "pilot mode" for the frame if time allows.

kimberscott:
Actually, the behavior sounds like it's already perfect! Upon clicking the "Return to Lookit" button is good. I just wanted to check it wasn't happening immediately upon checking the box. Thanks so much!

abought:
> Actually, the behavior sounds like it's already perfect! Upon clicking the "Return to Lookit" button is good. I just wanted to check it wasn't happening immediately upon checking the box. Thanks so much!

Durgh, that's what I get for context switching. :) I agree that warning someone right when they click a checkbox would be distracting, so we'll keep it as is. Good save.

abought:
Changes made! I have update the `exp-exit-survey-pilot` frame to be more consistent with changes made to the parent; we can further consolidate redundant code in the future.

I've also separated layout/readability changes from code changes in separate commits, to make the PR easier to read.

screen shot 2016-10-13 at 1 52 45 pm

PR #168, opened 2016-10-12
Update feature flags

samanehsan:
Companion to: https://github.com/CenterForOpenScience/isp/pull/51

## Purpose
The feature flags were refactored in this [commit](https://github.com/CenterForOpenScience/isp/pull/51/commits/53adda78ee736ead0b7e76a4a5ae7c214f191d48) and need to be updated where they are used as well.

### DIFF
--- a/exp-player/addon/components/exp-card-sort/component.js
+++ b/exp-player/addon/components/exp-card-sort/component.js
@@ -171,7 +171,7 @@ export default ExpFrameBaseComponent.extend({
     }).volatile(),
 
     allowNext: Ember.computed('cards.[]', function() {
-        if (config.validate) {
+        if (config.featureFlags.validate) {
             return this.get('cards').length === 0;
         }
         return true;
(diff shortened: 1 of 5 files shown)


exp-player/addon/components/exp-card-sort/component.js
```diff
@@ -171,7 +171,7 @@ export default ExpFrameBaseComponent.extend({
     }).volatile(),
 
     allowNext: Ember.computed('cards.[]', function() {
-        if (config.validate) {
+        if (config.featureFlags.validate) {
```
abought: This will work for us now.

It occurs to me that the flags are used in an addon, but defined in a consuming app. This creates a situation where these frames will *only* work if the consuming app happens to specify some undocumented config variables needed by the addon.... but since these frames are very narrowly special-purpose built for ISP, this is ok for now.

We're discussing some possible changes to how experimenter would work, but until that's clarified it's hard to advise on more specific changes. For now, this is just a general comment, and something to think about if we see bugs later.

PR #169, opened 2016-10-14
[rfr] Flags and timing events for various kinds of early exit

abought:
Refs: https://openscience.atlassian.net/browse/LEI-248
Companion to: TBD

# Purpose
If a user leaves a study early, capture event information.

# Summary of changes
- Add some default feature flag settings to `config()` for addon, so that non-ISP apps don't break when trying to consume code intended for ISP.
- Add flags/timing events for early exit so a record is saved in the session (see testing notes)
  - When user hits F1 key
  - When user leaves the page manually via back button, eg beforeunload events
- Minor code reformatting and cleanup

# Testing notes
## Sample of expected data
This adds a new field, `globalEventTimings`, to the attributes section of a session payload.

Any attempt to exit the study early should yield an `exitEarly` eventType. The type of the exit can be either:
- *browserNavigationAttempt*: Back button, close window, etc.
- *manualInterrupt*: The user elects to interrupt the study manually (eg by pressing the F1 key), such that they will be taken to the exit survey screen before leaving.

```
    "attributes": {
      ...
      "globalEventTimings": [
        {
          "exitType": "browserNavigationAttempt",
          "eventType": "exitEarly",
          "timestamp": "2016-10-14T18:42:04.341Z",
          "lastPageSeen": 7
        },
        {
          "exitType": "manualInterrupt",
          "eventType": "exitEarly",
          "timestamp": "2016-10-14T18:42:12.225Z",
          "lastPageSeen": 7
        }
      ],
    ...
```

## Testing procedure/ behavior
To test this, do various things, then look at the session payload generated at each step. (send a GET request with appropriate token to the session record URL of the form `https://the-server.com/v1/id/documents/experimenter.session<exptId>.<sessionId>`). You can get that URL by looking at the network tab- during an experiment, the browser sends PATCH requests to this URL.

1.  Participate in an experiment. At the consent screen (before recording a video), click the back button to navigate away.
  - Click "stay" to stay on the page. A `browserNavigationAttempt` exitEarly event should be logged.
  - Click the back button again, and then click "leave". A second `browserNavigationAttempt` will be logged with a more recent timestamp, and then you will actually be taken away from the page. 
2. Do the above- but instead of clicking the back button, quit the entire browser instead. A `browserNavigationAttempt` should be logged.
3. Participate in another experiment. At the consent screen, press `F1` (or `fn+F1` on a Mac, as appropriate). Whether you exit before or after filling out the survey, an exitType `manualInterrupt` should be captured.
4. Participate in an experiment. Attempt to leave, stay, then press F1. You should get one event of each exit type, the newest (successful) coming last.
5. Try the above in both Chrome and Firefox

## Caveats
Browser makers are making it progressively harder to interrupt any attempt to leave the page. (because of malicious sites that try to abuse the power to keep people there) We log the *attempt* to leave because we can not always determine whether the attempt is successful.

The logging event is done synchronously- the page should not be able to leave until the save is complete. It is possible, though unlikely, that a browser could decide to interrupt even synchronous usages in the future.

### DIFF
--- a/exp-models/addon/models/session.js
+++ b/exp-models/addon/models/session.js
@@ -13,6 +13,10 @@ export default DS.Model.extend(AnonJamModel, {
     conditions: DS.attr(),
     expData: DS.attr(),  // Data is a reserved keyword in ember
 
+    // Timing information for events captured by the player but not tied to a specific frame. Eg, early exit
+    //  from experiment
+    globalEventTimings: DS.attr({ defaultValue: () => [] }),
+
     profileId: DS.attr('string'), // Store ID of related record
 
     experimentId: DS.attr('string'),
--- a/exp-player/addon/components/exp-player.js
+++ b/exp-player/addon/components/exp-player.js
@@ -35,10 +35,11 @@ export default Ember.Component.extend(FullScreen, {
                 // There is no guarantee that the server request to save this event will finish before exit completed;
                 //   we are limited in our ability to prevent willful exits
                 this.send('setGlobalTimeEvent', 'exitEarly', {
-                    exitType: 'browserNavigation',
+                    exitType: 'browserNavigationAttempt', // Page navigation, closed browser, etc
                     lastPageSeen: this.get('frameIndex') + 1
                 });
-                this.get('session').save();
+                //Ensure sync - try to force save to finish before exit
+                Ember.run(() => this.get('session').save());
 
                 // Then attempt to warn the user and exit
                 let toast = this.get('toast');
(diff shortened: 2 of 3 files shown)


samanehsan:
I'm testing this behavior in ISP and I do not get a message when I click the back button (perhaps because I handle the back button to go to the previous frame/page).

As expected, when I exit the browser a `browserNavigationAttempt` is logged and when I skip to the last frame by clicking F1, a `manualInterrupt` is logged. As discussed in person, participants should not be able to skip the survey by clicking F1 (but this can be addressed in a follow-up PR)

PR #171, opened 2016-10-17
Refactor player to extract Lookit-specific exit behaviors

abought:
Refs: https://openscience.atlassian.net/browse/LEI-267
Companion to: 

## Purpose
The player currently has certain custom behaviors that are Lookit-specific. We do not, for example, want ISP users to be able to skip to the last frame by pressing F1.

Refactor to move those behaviors inside lookit.

## Summary of changes
- Reorg code: move lookit-only functionality into Lookit.
- Add a new `beforeUnload` hook to the player, controlling additional custom behavior that happens when a user leaves. For example, it can be used to display a toast message with more info
- Add a new `messageEarlyExitModal` property, controlling text of the confirm-on-exit modal. (may not show this text in all browser, but method has to return a non-null value if you want the modal to appear)

## Testing notes
- All studies will show warning messages with unsaved data by default 
- ISP will not respond to the F1 key
- Lookit will respond to the F1 key by skipping ahead to the last frame

### DIFF
--- a/exp-player/addon/components/exp-player.js
+++ b/exp-player/addon/components/exp-player.js
@@ -8,6 +8,26 @@ let {
     $
 } = Ember;
 
+/**
+ * @module exp-player
+ * @submodule components
+ */
+
+/**
+ * Experiment player: a component that renders a series of frames that define an experiment
+ *
+ * Sample usage:
+ * ```handlebars
+ * {{exp-player
+ *   experiment=experiment
+ *   session=session
+ *   pastSessions=pastSessions
+ *   saveHandler=(action 'saveSession')
+ *   frameIndex=0
+ *   fullScreenElementId='expContainer'}}
+ * ```
+ * @class exp-player
+ */
 export default Ember.Component.extend(FullScreen, {
     layout: layout,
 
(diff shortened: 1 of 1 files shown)


samanehsan:
Confirmed that ISP no longer responds to the F1 key. Also, a warning message does show up when attempting to close the browser or navigate to a different url, but this happens regardless of whether there is unsaved data. Is that intentional @abought?

abought:
> Confirmed that ISP no longer responds to the F1 key. Also, a warning message does show up when attempting to close the browser or navigate to a different url, but this happens regardless of whether there is unsaved data. Is that intentional @abought?

In theory, the warning message should always have shown up- the removed code simply depended on a boolean property, which would have had the same value for both players:

https://github.com/CenterForOpenScience/exp-addons/pull/171/files#diff-85b6be66a5ce62f39ee05c9d04ceb67fL30

Will follow up.

abought:
As discussed on flowdock: the reason that ISP is suddenly getting an error message is because I moved the `toast.warning` call into lookit.

Although the player code used toast, neither the player nor ISP listed it as a dependency. Attempting to call a toast method from ISP was, before this PR, causing the `beforeunload` code to fail halfway through.

samanehsan:
As @abought pointed out on flowdock, the message shows up depending on the `allowExit` variable: 
https://github.com/CenterForOpenScience/exp-addons/pull/171/files#diff-85b6be66a5ce62f39ee05c9d04ceb67fR46
https://github.com/CenterForOpenScience/exp-addons/pull/171/files#diff-85b6be66a5ce62f39ee05c9d04ceb67fR62

It does not depend on unsaved data so the current behavior is correct. That was my only concern, so this is good to go 🎉

PR #172, opened 2016-10-17
Remove experiment conditions [LEI-262]

samanehsan:
Companion to: 

## Purpose
Replace the two different experiment conditions (describing a situation at 10am or 7pm) with a question asking the participant what time the situation occurred.

## Summary of changes
- Ask the participant to describe a memorable situation and what time it happened: 
screen shot 2016-10-17 at 5 48 50 pm
- Use moment.js to localize times for the drop-down question "Approximately what time did the experience begin?" The `longDateFormat` key for unsupported locales are defined in `config/environment.js`
- Exp-rating form question 3 now asks "Please rate your behavior in the situation you described."

## Testing notes
- Select English(US) from the flag-picker modal. On the second survey page, ensure the options for the first question "Approximately what time did the experience begin?" are in 12-hr AM/PM notation. 
- Select English(GB) from the flag-picker modal. For the same question as before, ensure the times are using 24-hour notation.

### DIFF
--- a/exp-player/addon/components/exp-free-response/component.js
+++ b/exp-player/addon/components/exp-free-response/component.js
@@ -50,9 +50,13 @@ export default ExpFrameBaseComponent.extend(Validations, {
     }),
     times: Ember.computed(function() {
         var times = [];
-        for (var i=0; i < 24; i++) {
+        for (var i=0; i <= 24; i++) {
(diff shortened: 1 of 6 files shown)


samanehsan:
Replaced with https://github.com/CenterForOpenScience/exp-addons/pull/174

PR #174, opened 2016-10-18
Feature: Add question asking what time situation occurred [LEI-262]

samanehsan:

## Purpose
On the exp-free-response component, add question asking the participant what time the situation occurred.

## Summary of changes
- Ask the participant to describe a memorable situation and what time it happened: 
screen shot 2016-10-17 at 5 48 50 pm
- Use moment.js to localize times for the drop-down question "Approximately what time did the experience begin?" The `longDateFormat` key for unsupported locales are defined in `config/environment.js`

## Testing notes
- Select English(US) from the flag-picker modal. On the second survey page, ensure the options for the first question "Approximately what time did the experience begin?" are in 12-hr AM/PM notation. 
- Select English(GB) from the flag-picker modal. For the same question as before, ensure the times are using 24-hour notation.

### DIFF
--- a/exp-player/addon/components/exp-free-response/component.js
+++ b/exp-player/addon/components/exp-free-response/component.js
@@ -50,9 +50,13 @@ export default ExpFrameBaseComponent.extend(Validations, {
     }),
     times: Ember.computed(function() {
         var times = [];
-        for (var i=0; i < 24; i++) {
+        for (var i=0; i <= 24; i++) {
             var date = new Date(2016, 10, 17, i, 0, 0);
-            // TODO: Handle missing moment locales: amharic, urdu, runyakore, luganda,             longDateFormat : {
+            var locale = this.get('i18n.locale');
+            if (locale in config.localeTimeFormats) {
+                // Add locale information not specified by moment.js
+                moment.updateLocale(locale, config.localeTimeFormats[locale])
+            }
             moment.locale(this.get('i18n.locale'));
             times.push(moment(date).format('LT'));
         }
--- a/exp-player/addon/components/exp-free-response/component.js
+++ b/exp-player/addon/components/exp-free-response/component.js
@@ -3,6 +3,7 @@ import ExpFrameBaseComponent from 'exp-player/components/exp-frame-base';
 import layout from './template';
 import {validator, buildValidations} from 'ember-cp-validations';
 import config from 'ember-get-config';
+import moment from 'moment';
 
 
 function getLength(value) {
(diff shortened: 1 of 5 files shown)


exp-player/config/environment.js
```diff
@@ -2,5 +2,32 @@
 'use strict';
 
 module.exports = function(/* environment, appConfig */) {
-  return { };
+  return {
+      moment: {
+          includeLocales: true
```
abought: For any front-end application, we should try to avoid having the user download more information than is needed.

Does this option cause *all* locales to be part of the final built app? As an easy test: if you tell ember-moment to only use en-us, is the file size from `ember build` the same size, or smaller, than if you use `includeLocales: true` (the "everything" option)?

https://github.com/stefanpenner/ember-moment#i18n-support

samanehsan: Good point! I can change that to include just the locales ISP needs.

exp-player/addon/components/exp-free-response/component.js
```diff
@@ -46,12 +48,28 @@ export default ExpFrameBaseComponent.extend(Validations, {
         message = message.replace("0", length.toString());
         return message;
     }),
+    times: Ember.computed(function() {
+        var times = [];
+        for (var i=0; i <= 24; i++) {
+            var date = new Date(2016, 10, 17, i, 0, 0);
```
abought: Do you mind if I ask why this hard-codes Monday's date specifically?

exp-player/addon/components/select-input/component.js
```diff
@@ -5,5 +5,6 @@ import layout from './template';
 export default Ember.Component.extend({
   layout,
   options: null,
-  value: null
+  value: null,
+  translate: true
```
abought: For future development, I am trying to encourage that all components document the kinds of parameters they expect to be passed in, as well as what those parameters are for.

We're using YUIDoc syntax for this. Here is an example: https://github.com/CenterForOpenScience/ember-osf/blob/develop/addon/components/osf-navbar/component.js#L34

One caveat: in the example above, you'll also notice that it includes some boilerplate ndeclaring it as a class at the top of the file. Without that, yuidoc doesn't actually detect what it should be documenting.
https://github.com/CenterForOpenScience/ember-osf/blob/develop/addon/components/osf-navbar/component.js#L10

Let me know if I can help you get started with the documentation process. But for now, at very least the reusable component should tell you how to use it!

exp-player/config/environment.js
abought: I see this was moved back to code review, but this comment was unaddressed. Can you comment on the decision to leave as is? Are there technical barriers I can help with?

PR #176, opened 2016-10-18
Add labels to overview form radio questions [LEI-263]

samanehsan:
Refs: https://openscience.atlassian.net/browse/LEI-263

## Purpose
- Add labels to the radio-button questions on the `exp-overview` form
- Add "I prefer not to answer" option to "On a scale from 1-10, how religious are you?"

## Changes
screen shot 2016-10-19 at 11 12 07 am

### DIFF
--- a/exp-player/addon/components/exp-overview/component.js
+++ b/exp-player/addon/components/exp-overview/component.js
@@ -62,7 +62,7 @@ const questions = [
     type: 'radio',
     scale: range(1, 10),
     labelTop: false,
-    formatLabel: 'negative-margin-top',
+    formatLabel: 'negative-margin-top narrow-width',
     value: null,
     labels: [
         {
@@ -107,7 +107,7 @@ const questions = [
     scale: range(1, 11),
     hiddenOptions: [11],
     labelTop: false,
-    formatLabel: 'negative-margin-top',
+    formatLabel: 'negative-margin-top narrow-width',
     value: null,
     labels: [
         {
--- a/exp-player/addon/components/exp-overview/component.js
+++ b/exp-player/addon/components/exp-overview/component.js
@@ -29,11 +29,6 @@ var generateValidators = function(questions) {
   return validators;
 };
 
-var getQuestion8Options = function() {
-    var scale = range(1, 10);
-    scale.push('survey.sections.1.questions.8.options.preferNoAnswer');
-    return scale;
-};
 
 const questions = [
   {
@@ -109,7 +104,8 @@ const questions = [
   {
     question: 'survey.sections.1.questions.8.label',
     type: 'radio',
-    scale: getQuestion8Options(),
+    scale: range(1, 11),
+    hiddenOptions: [11],
     labelTop: false,
     formatLabel: 'negative-margin-top',
     value: null,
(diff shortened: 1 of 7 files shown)


exp-player/addon/components/exp-overview/component.js
```diff
@@ -29,6 +29,12 @@ var generateValidators = function(questions) {
   return validators;
 };
 
+var getQuestion8Options = function() {
+    var scale = range(1, 10);
+    scale.push('survey.sections.1.questions.8.options.preferNoAnswer');
```
abought: How will this field be represented in the exported data?

For example, if they analyze it always expecting a number, what will their data analysis program do if it sees a string? Is there a more standardized way of representing missing data for their programs? If so, should we consider storing the response that way instead of storing a string?

samanehsan: Good question! According to their code book, "I prefer not to answer" is represented by the number 11.

abought: Hmm, might want to tweak that before merge then, so that no bad sample data makes it into production.

samanehsan:
Done making changes @abought!

exp-player/addon/components/isp-radio-group/template.hbs
```diff
@@ -2,21 +2,37 @@
...
               {{#if (eq value.rating option)}}
-                {{t value.label}}
+                <span class="{{if value.formatClass value.formatClass}}">
+                    {{t value.label}}
+                </span>
               {{/if}}
             {{/each-in}}
             {{radio-button value=(last-segment option) checked=value}}
-            {{t option}}
+            {{#if hiddenOptions}}
```
abought: This code renders spans that contain the label, but with a class to mark it as hidden.

If you don't want the label to be visible, wouldn't it be simpler to not render the label at all in this block? Line 31 is sending sort of a mixed message.

exp-player/addon/styles/components/exp-overview.scss
```diff
@@ -0,0 +1,5 @@
+.format-average-label {
```
abought: I continue to be worried about all the special-case styles for each individual element- especially since these are pixel-specific placements.

What happens when you change the default font size, zoom the page, resize the browser, or translate this string to a longer/shorter word? Does the label still appear centered in the way you expect?

exp-player/addon/components/isp-radio-group/template.hbs
samanehsan: That's what I did initially, but if the span isn't there it throws off the alignment of the radio buttons

exp-player/addon/styles/components/exp-overview.scss
abought: Table cell widths are more a suggestion, and text may still overflow positions if a single word (without spaces) is unusually long. At some point I really strongly encourage you to try a few likely situations outside the normal as outlined- not just making it good in english on a big monitor, but deliberately trying to break the assumptions behind your layout and seeing what happens.

In particular, I'm worried that we have some table cells that break if a word or span is missing. That seems like it could be very sensitive to word lengths.

exp-player/addon/components/isp-radio-group/template.hbs
samanehsan: The vertical alignment depends on the height, so if the option wraps onto a second line the radio button gets pushed up and if there is no option label the radio button gets pushed down. It's not an issue for this particular question since the number options will not be translated. But it could definitely be an issue for questions where the options are words and not numbers.

PR #178, opened 2016-10-20
Add question asking time of situation [LEI-262]

samanehsan:
Companion to: https://github.com/CenterForOpenScience/isp/pull/89

## Purpose
Add question asking participants approximately what time the situation they are describing began.

## Summary of changes
Use ember-power-select component for the drop-down to achieve the desired styling.

### DIFF
--- a/exp-player/addon/components/exp-free-response/component.js
+++ b/exp-player/addon/components/exp-free-response/component.js
@@ -1,4 +1,3 @@
-var $ = Ember.$;
 import Ember from 'ember';
 import ExpFrameBaseComponent from 'exp-player/components/exp-frame-base';
 import layout from './template';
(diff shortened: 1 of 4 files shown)


exp-player/addon/components/exp-free-response/component.js
```diff
@@ -1,3 +1,4 @@
+var $ = Ember.$;
```
abought: This symbol appears to be unused- can it be removed?

exp-player/addon/components/exp-free-response/template.hbs
```diff
@@ -2,6 +2,23 @@
   <p class="break-line">{{t 'survey.sections.2.instructions.firstSection'}}</p>
   {{#bs-form}}
     {{#bs-form-group}}
+      <label>{{t 'survey.sections.2.questions.14.label'}}</label><br>
+        {{#power-select
+          selected=time
+          options=times
+          placeholder=placeholder
+          searchEnabled=false
+          onchange=(action (mut time))
```
abought: Can you clarify how the pick-a-time question ties together?

1. The field doesn’t appear to be tied to a response field. Eg, it iterates over `times as time`, but when you select, it mutates `time`, rather than a value that could be tied to a specific response...
https://github.com/CenterForOpenScience/exp-addons/pull/178/files#diff-04664267214cdc73b1c7ad09d204cf3eR11
2. The answer field *seems* to be `responses.q4`, if just going by a new field is a guide, but its label is `survey.sections.2.questions.14.label`. This nomenclature seems confusing.
3. You have two time values in the select box- which representation actually gets persisted to the database? (we don’t want to be sending the entire 12/24hr JSON object to the DB field)

samanehsan:
Done fixing the data serialization & CSS @abought! ⛅️

exp-player/package.json
```diff
@@ -57,7 +57,8 @@
     "broccoli-merge-trees": "^1.1.1",
     "ember-cli-babel": "^5.1.5",
     "ember-cli-htmlbars": "0.7.4",
-    "ember-cp-validations": "2.9.2"
+    "ember-cp-validations": "2.9.2",
```
abought: The current versions are 2.9.7 (point release) or 3.1.0. Can you please clarify why this PR pins to an older version instead?

exp-player/package.json
```diff
@@ -57,7 +57,8 @@
...
+    "ember-power-select": "0.10.11"
```
abought: This pins to a different version than the companion PR, which is mildly confusing.
https://github.com/CenterForOpenScience/isp/pull/89/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R50

Once that's resolved clear to merge.

samanehsan: Fixed the versions @abought!

PR #187, opened 2016-10-28
Fix CSS for long translations [LEI-282]

samanehsan:
Refs: https://openscience.atlassian.net/browse/LEI-282
## Purpose

Check the formatting of labels, questions, etc. using a pseudo-translation (double each translation string) to identify places where the UI may look off with long translation strings.
## Summary of changes
- Wrap bucket labels that are too long so that they do not overlap another label
- Put bucket labels into a separate row so that the buckets can stay aligned even
  when the labels are of differing lengths
- Hide the label row on xs screens, because the row will collapse and list all of the labels and
  then all of the buckets.
- Show the bucket labels inside the bucket row on xs screens so that the labels appear directly   above the bucket.
- Set a max-width for isp-radio-component buttons so that the text wraps onto multiple lines instead of increasing the width of the label element

### DIFF
--- a/exp-player/addon/components/exp-card-sort/template.hbs
+++ b/exp-player/addon/components/exp-card-sort/template.hbs
@@ -19,10 +19,17 @@
   </div>
   <h4 class="center">{{cards.length}} {{t 'qsort.sections.1.itemsLeft'}}</h4>
 {{/if}}
+<div class="row hidden-xs">
+  {{#each buckets as |bucket|}}
+    <div class="col-sm-4">
+      <h3 class="bucket-label" style={{height}}>{{t bucket.name}}</h3>
+    </div>
+  {{/each}}
+</div>
 <div class="row">
     {{#each buckets as |bucket|}}
       <div class="col-sm-4">
-          <h3 class="text-center">{{t bucket.name}}</h3>
+          <h3 class="bucket-label visible-xs" style={{height}}>{{t bucket.name}}</h3>
           {{#if (eq framePage 0)}}
             {{#draggable-object-target bucket=bucket.cards cards=cards buckets=buckets action='dragCard'}}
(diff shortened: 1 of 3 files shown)


exp-player/addon/components/exp-card-sort/template.hbs
```diff
@@ -19,10 +19,17 @@
...
+    <div class="col-sm-4">
+      <h3 class="bucket-label" style={{height}}>{{t bucket.name}}</h3>
+    </div>
+  {{/each}}
+</div>
 <div class="row">
     {{#each buckets as |bucket|}}
       <div class="col-sm-4">
-          <h3 class="text-center">{{t bucket.name}}</h3>
+          <h3 class="bucket-label visible-xs" style={{height}}>{{t bucket.name}}</h3>
```
abought: For future note: Why is this a heading? Headings imply a structure to the document, but this appears to be just a label.

We should generally avoid using heading tags for pure formatting; it confuses screen readers and makes it harder for vision-impaired users to use the site.

exp-player/addon/components/exp-card-sort/template.hbs
```diff
@@ -52,7 +59,20 @@
         <div class="row">
           {{#each group.categories as |category|}}
             <div class="col-sm-4 small-bucket">
-              <h5 class="text-center {{if (eq category.name 'qsort.sections.2.categories.neutral') '' 'bucket-label'}}">{{t category.name}}</h5>
+              <h5 class="bucket-label hidden-xs {{if (eq category.name 'qsort.sections.2.categories.neutral') '' 'bucket-label-margin'}}">{{t category.name}}</h5>
```
abought: See general note above: we should avoid using headings for cosmetic purposes. (also line 75 below)

PR #189, opened 2016-10-29
Upgrade exp-addons to ember 2.8 LTS

abought:
Refs: https://openscience.atlassian.net/browse/LEI-294
Companion to: TBD

⚠️ This will require coordinated verification with experimenter, LEI, and ISP before the platform upgrade can be considered complete. 
## Purpose

Upgrade exp-addons projects to use ember-2.8 and resolve any relevant deprecation warnings observed.
## Summary of changes
- Upgrade to ember 2.8
  - [x] exp-models
  - [x] exp-player
## Testing notes

All projects on the experimenter platform will have to be validated for the upgrade. Changing ember versions by 10 months is not to be taken lightly and we should re-qa all site functionality to ensure it works as expected.


abought:
Verified working with experimenter and ISP.

Will merge this as is to support ISP QA tomorrow. If Lookit has further issues that require attention, will fix in a followup PR.

PR #193, opened 2016-11-01
Clear unsaved cardsort1 data [LEI-302]

samanehsan:
Refs: https://openscience.atlassian.net/browse/LEI-302

## Purpose
If a user sorts cards on the cardsort1 page, then clicks the "Previous Page" button to
go to the previous page, and then clicks the "Continue" button to go back to the cardsort1 page, they should see 90 cards at the top to be sorted and three empty buckets to sort them into. Instead, the buckets still contain the previously sorted cards. 

## Summary of changes
When a user clicks the "Previous Page" button on the cardsort1 page, clear the sorted cards.

## Testing notes
Steps to reproduce: 
1. On the first cardsort page, sort at least one card. Notice the number of cards left decreases. 
2. Click the "Previous Page" button
3. Click the "Continue" button to return to the cardsort page
4. Notice that the number of cards left is 90, but the sorted cards still re-appear.

### DIFF
--- a/exp-player/addon/components/exp-card-sort/component.js
+++ b/exp-player/addon/components/exp-card-sort/component.js
@@ -239,6 +239,13 @@ export default ExpFrameBaseComponent.extend({
                 window.scrollTo(0, 0);
             }
         },
+        previousPage() {
+            // clear unsaved data
+            for (let bucket of this.get('buckets')) {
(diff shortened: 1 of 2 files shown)


abought:
Idea looks sound; I'll trust that Saman did functional review on this one and move it forward for QA.

PR #197, opened 2016-11-06
Fix card text overflowing card

samanehsan:
Refs: https://trello.com/c/e5LbqRT8/4-text-on-cards-in-sub-categories-runs-off-card

## Purpose
On small screen widths, card text runs off of cards.

## Summary of changes
- Make buckets wider
- Make cards wider
- Make card text break in between long words 

screen shot 2016-11-06 at 6 44 08 pm  

screen shot 2016-11-06 at 6 44 20 pm

### DIFF
--- a/exp-player/addon/styles/components/exp-card-sort.scss
+++ b/exp-player/addon/styles/components/exp-card-sort.scss
@@ -34,6 +34,7 @@
     padding: 10px;
     margin-bottom: 12px;
     overflow-wrap: break-word;
+    word-wrap: break-word; // support for original name: https://developer.mozilla.org/en-US/docs/Web/CSS/overflow-wrap
 }
 
 .bucket-label-margin {
--- a/exp-player/addon/styles/components/exp-card-sort.scss
+++ b/exp-player/addon/styles/components/exp-card-sort.scss
@@ -21,13 +21,19 @@
     font-size: 10px;
 }
 
+.small-bucket {
+    padding-right: 8px;
+    padding-left: 8px;
+}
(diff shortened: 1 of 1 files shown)


abought:
Looks like purely cosmetic changes here; no review needed.

This will look weird on small screens, but ultimately there's not much we can do if the column is narrower than the text.

exp-player/addon/styles/components/exp-card-sort.scss
```diff
@@ -21,13 +21,19 @@
...
+
 .small-bucket .well {
-    padding:12px;
+    padding:10px;
 }
 
 .small-bucket .alert {
     padding: 10px;
     margin-bottom: 12px;
+    overflow-wrap: break-word;
```
samanehsan: Currently investigating why `overflow-wrap` does not break up long words in firefox

abought: Might want to check browser version. According to the MDN browser compatibility table for this property, support was added in... FF 49? Seriously?

abought: http://caniuse.com/#search=overflow-wrap

https://developer.mozilla.org/en-US/docs/Web/CSS/overflow-wrap#Browser_compatibility

samanehsan: As discussed via flowdock, I was using Firefox 44, which does not support `overflow-wrap`, so I also specified `word-wrap` (the previous name for the property)

PR #201, opened 2016-11-22
Update code for final version of study! 

kimberscott:
Refs: https://openscience.atlassian.net/browse/LEI-
Companion to: 

## Purpose

Get ready to test for physics study!

## Summary of changes

Update pref-phys randomizer to use final versions of stimuli and new counterbalancing procedure.

Minor wording changes.

## Testing notes

'Physics - final version draft' (5834bc793de08a003a5dfaa9) on staging has a draft of what the actual final experiment should look like.

pref-phys-test recreates appropriate tests from pref-phys-pilot-test, but doesn't currently do anything further. (Some crude tests are used for local development, using an HTML page that displays the conditions & list of videos to show, but that requires not loading Ember or exporting anything in pref-phys.) 

Study works well for Kim running Lookit locally and using Chrome. I can't load the site *at all* in Firefox (localhost:4200 shows the site title and icon, but is otherwise empty; lookit.mit.edu and staging-lookit.osf.io are fine), which I am hoping is not due to anything I've changed here.


exp-player/addon/components/exp-frame-base/component.js
```diff
@@ -139,7 +139,7 @@ export default Ember.Component.extend({
             this.send('setTimeEvent', 'nextFrame');
             this.send('save');
             this.sendAction('next');
-            window.scrollTo(0, 0);
+            //window.scrollTo(0, 0);
```
abought: Question: Can you please comment on why this was turned off?

`exp-frame-base` is used by all experiments. A quirk of ember is that by default, pages don't automatically scroll back to the top when you change the page (next, previous, route transitions, etc). `window.scrollTo(0,0)` is intended to give people a more normal page change experience, and turning it off in the base frame would affect experiments by other groups.

Perhaps I can help identify a good alternative?

exp-player/addon/components/exp-physics-intro/template.hbs
```diff
@@ -30,12 +28,6 @@
             select a privacy level for your webcam recording.
         </p>
         <br>
-        <p style="color:blue;">
```
abought: With the removal of the text, is it safe to also clean up the other sections dealing with confirmation code?

https://github.com/CenterForOpenScience/exp-addons/pull/141/commits/6bcc13378ab3b475a3234b56a02478fca875d625

(We try to remove dead code before we forget about it)

exp-player/addon/randomizers/pref-phys.js
```diff
@@ -10,105 +12,1336 @@ function shuffleArray(array) {
...
+        ['cup', 'orangeball', 'whiteball', 'apple', 'spray', 'lotion'],
+        ['apple', 'whiteball', 'lotion', 'spray', 'cup', 'orangeball'],
+        ['orangeball', 'cup', 'whiteball', 'spray', 'apple', 'lotion'],
+        ['orangeball', 'apple', 'cup', 'lotion', 'whiteball', 'spray'],
+        ['cup', 'apple', 'lotion', 'spray', 'whiteball', 'orangeball'],
+        ['orangeball', 'apple', 'whiteball', 'spray', 'cup', 'lotion'],
+        ['cup', 'orangeball', 'whiteball', 'lotion', 'spray', 'apple'],
+        ['orangeball', 'whiteball', 'apple', 'cup', 'lotion', 'spray'],
+        ['whiteball', 'lotion', 'orangeball', 'apple', 'spray', 'cup'],
+        ['whiteball', 'lotion', 'orangeball', 'apple', 'cup', 'spray']
```
abought: One quirk of front-end development is that each line of code we write has to be downloaded by the browser. To help both maintainability and download speed, are there any reasons not to consolidate this down to a single line that generates every possible combination, just given the choices?

That seems like a good fit for `conceptOrderRotation` and `supportObjectRotation`; it's hard to tell whether `gravityObjectRotation` contains quite every possible combination (there don't seem to be enough lines, unless I miscounted the number of items).

A sample would be something like the below:
```
function pseudocodeThatShuffles(array) {
    // Do stuff
    return shuffled;
}

const concepts = ['control', 'inertia', 'support', 'gravity'];

let conceptOrderRotation = pseudoCodeThatShuffles(concepts);

```

exp-player/addon/randomizers/pref-phys.js
```diff
@@ -10,105 +12,1336 @@ function shuffleArray(array) {
...
+        ['hammer', 'book', 'brush', 'duck', 'tissues', 'shoe'],
+        ['hammer', 'shoe', 'tissues', 'book', 'duck', 'brush'],
+        ['book', 'brush', 'tissues', 'duck', 'shoe', 'hammer'],
+        ['book', 'brush', 'duck', 'shoe', 'hammer', 'tissues'],
+        ['shoe', 'duck', 'hammer', 'book', 'tissues', 'brush'],
+        ['hammer', 'tissues', 'duck', 'brush', 'shoe', 'book'],
+        ['duck', 'book', 'brush', 'shoe', 'tissues', 'hammer'],
+        ['duck', 'shoe', 'book', 'hammer', 'tissues', 'brush'],
+        ['shoe', 'brush', 'hammer', 'duck', 'tissues', 'book'],
+        ['brush', 'tissues
```
abought: We try to minimize console.log statements and often remove local debugging code before merging; do we want to keep this statement in? 

Also, if this is for more than just local debugging: Is this meant to be a likely and tolerable error that we can ignore, or is this an event that we explicitly want to see logged to sentry?

exp-player/addon/randomizers/pref-phys.js
```diff
@@ -206,65 +1450,33 @@ function assignVideos(startType, showStay, whichObjects, NPERTYPE) {
...
+            var altName;
+            if (e.compType === 'calibration') {
+                fname = `sbs_calibration_${e.flip}`;
+                filenames.push(fname);
+                altName = `sbs_calibration_${e.flip}`;
+                e.fname = fname;
+                e.altName = altName;
+            } else {
+                fname = `sbs_${e.compType}_${e.outcomeL}_${e.outcomeR}_${e.object}_${e.camera}_${e.background}_${e.flip}`;
+                filenames.push(fname);
```
abought: Incidentally, this is a truly extraordinary number of videos! I know we saw a few missing videos early on in piloting; amazing to see how many must have come along since.

abought:
I am able to load the site in Firefox 49.0.2, Mac OS X, but that's a very interesting bug report. Are there any red errors on the Developer Tools Javascript console that might help track down the problem?

(other than "raven is not configured", which is ok- that just means you're not sending messages to sentry from your local dev environment, and it's helpfully telling you so)

exp-player/addon/randomizers/pref-phys.js
abought: Also, on further review we don't want shuffling so much as permutations. (quick erratum)

exp-player/addon/components/exp-frame-base/component.js
abought: That sounds like a reset bug worth revisiting later, but reverting this change for now would help.

We can create a followup ticket for the issue you report, which may be the preferred way to target a fix so we don't forget to coordinate across projects after the holiday weekend.

exp-player/addon/randomizers/pref-phys.js
kimberscott: Ah, so this is a bit tricky. These lists are needed for counterbalancing to distribute object-comparison mappings *roughly* evenly across kids, and then to ensure (without requiring complicated queries about their individual histories) that those kids see different pairings as they keep participating. Cycling through a list of such mappings, in a fixed pseudorandom order, is one of the simpler and more transparent/reproducible ways to arrange this. 

It's critical here that the lists are in a **fixed** but (pseudo-)random order; they're of course programmatically generated, but only once. It would be possible to move the code that generates them to the randomizer, but we'd need to seed whatever random generator was used and be absolutely certain that changes to any underlying code could never affect the lists generated. (And yes, not all of them actually list all the permutations--gravity needs to leave out ones that pair 'table' with 'spray' or 'ramp' with 'orangeball', inertia and control have some redundancy removed because we're repeating the same comparison types twice each.) My best guess was that loading this bunch of text was worth it for better scientific transparency, since we're talking about a file size orders of magnitude less than the videos--does that sound reasonable?

Another option is to basically manually compress this data--there's of course no reason that these permutations have to explicitly list the strings over and over; we could represent each permutation with a number and then parse those out. But that would actually make it harder to see what's going on.

A third option would be to subsample from the list of possible permutations. However, scientifically it's better to be sampling from the whole distribution, which will necessarily reduce the variance in e.g. how many kids see 'whiteball' paired with 'stop hand/nohand' more often than with 'reverse barrier/nobarrier'. 

So basically yes, this is terrible, it can be moved to another file if that would be nicer to maintain, but we do have reasons for doing it.

* The list is in a **fixed**, but (pseudo-)random, order. (It defeats the purpose for the list to be shuffled upon 
*

kimberscott:
Nope, there were no JS errors whatsoever, and there *was* HTML being loaded, just not that made for anything showing up. Annnnd now I can't reproduce it, so who knows. I'm sorry!

exp-player/addon/randomizers/pref-phys.js
abought: How was the original list generated? I assume not by hand? Can we use the code you used for the best and most direct comparison?

Insofar as some of these lists were manually filtered (not 100% algorithmically determined), that does add complexity.

I'm tentatively willing to merge this now insofar as we plan to break lookit-specific code out of the shared codebase. (people running other experiments shouldn't need to download this list). The best path forward would be:

1. If you already have the function used to generate the list, then the lists of permutations that *can* be created using that function should be. (eg `conceptOrderRotation`). Most of the permutation-generating algorithms are deterministic and would generate the same order of arrangements every time.

2. We'll leave the "manually curated" lists of permutations as is, because this requires knowledge beyond a brute force algorithm.


You make a good point about clarity and reproducibility, but the tradeoff in those long hard-coded lists  is maintainability; if we use a list of 720 lines * 6 strings each, there's some concern that a single stray typo could change the results and be very hard to track down. 

I think a reliable procedure might be less vulnerable to errors than pure hard-coding. Would that be a viable solution?

Pure technical aside: If we are worried about the performance penalty of generating the permutations, there do seem to be ways to predictably get the n-th permutation. I just wouldn't go tacking them in untested an hour before a holiday release, is all. ;)
http://stackoverflow.com/a/7919887/1422268

exp-player/addon/components/exp-physics-intro/template.hbs
abought: > Nope, there were no JS errors whatsoever, and there was HTML being loaded, just not that made for anything showing up. Annnnd now I can't reproduce it, so who knows. I'm sorry!

I actually like when bugs go away on their own! The most common reasons for things to not show up are a build error (eg if you were missing a dependency that got picked up when you restart your ember server), or a slow server (ember pages that depend on JamDB wait to load until the data has arrived, and sometimes the internet gets sluggish for a few minutes). We will monitor for future issues then.

. But it is something to watch for.

exp-player/addon/randomizers/pref-phys.js
abought: Hmm. Mostly these are advisory notes; I can see where the more subtle requirements would add complexity to any nice procedural method. I recognize there's a time crunch here as well, and will leave the final decision to you.

Mainly, we'll just need to be extremely careful not to change anything about this randomizer later once the experiment starts. In that light, our caution is partly because we won't be able to QA and regression test this in depth ourselves; if you've had a chance to verify that counterbalancing is working as expected, then that helps.

kimberscott: Okay, let's just go ahead with the hard-coded lists. I do appreciate the caution, though.

We'd actually have needed to be extremely careful not to change anything about this randomizer regardless; part of my concern about procedurally generating the fixed pseudo-random lists on the fly is that it wouldn't be as glaringly obvious that something had subtly changed (even though changing the random seed, or changing whether the state updated between different lists, or changing the order in which the lists were generated, etc. would violate the expected behavior just as badly as introducing a new 720-line list).

abought:
Code merged, and I've confirmed it has deployed successfully to https://staging-lookit.osf.io . Note for record-keeping purposes: as discussed with Kim ("show outdated" collapsed comments on this thread), MIT will take point on verifying the randomizer and making sure that this passes any required manual regression testing. The long hard-coded lines are a short-term constraint due to complex requirements for permutations.

The only change we will plan to make to this code will involve a future move of Lookit-specific code into the Lookit app (so that ISP users are not required to download these lists in their SPA). This will most likely happen sometime after the new year.

Here are some notes on recent randomizer-specific regression testing that may be helpful, though of course I assume Kim has found these already. :)

https://github.com/CenterForOpenScience/exp-addons/pull/150
https://github.com/CenterForOpenScience/exp-addons/pull/158

Sadly I will only have time today to do light general regression testing of video playback, and our target is to request deployment to production no later than 2pm today so we have time to respond if the deployment server hiccups. Let me know if slight  adjustments to the schedule would help!

kimberscott:
Thanks, & sounds good! Yes, we'll handle checking the randomizer behavior (beyond the unit tests).

abought:
Thanks again! On the assumption that no news is good news, I'll plan to make the `Lookit 1.0` release branch ~1:45 pm today if we don't hear anything, and send it to DevOps at 2pm for them to put on the production server.

PR #205, opened 2016-12-07
Add csv-writer

binoculars:
Refs: https://openscience.atlassian.net/browse/LEI-246

## Purpose
Utility function for generating a CSV file from an array of objects

## Summary of changes
* Add csv-writer utility with tests
* Update travis and README

## Testing notes
Run `npm link` in exp-models to ensure that the module is able to be referenced by name (instead of by path)

### DIFF
--- a/exp-models/addon/utils/csv-writer.js
+++ b/exp-models/addon/utils/csv-writer.js
@@ -17,7 +17,7 @@ export const NL = '\r\n';
 function squashPrefix(obj, prefix) {
     let ret = {};
 
-    for (const key in Object.keys(obj)) {
+    for (const key of Object.keys(obj)) {
         const newPrefix = prefix ? `${prefix}.${key}` : key;
 
         let value = Ember.get(obj, key);
--- a/exp-models/addon/utils/csv-writer.js
+++ b/exp-models/addon/utils/csv-writer.js
@@ -1,0 +1,107 @@
+import Ember from 'ember';
+
+/**
+ * New line string
+ *
+ * @type {string}
+ */
+export const NL = '\r\n';
+
+/**
+ * Flatten a nested object into a single level, with dotted paths for keys
+ *
+ * @param obj {!Object} - The object to flatten
+ * @param prefix {string} - The prefix for the keys
+ * @returns {Object} - The flattened object
+ */
+function squashPrefix(obj, prefix) {
+    let ret = {};
+
+    for (const key in obj) {
+        if (!obj.hasOwnProperty(key)) {
+            continue;
+        }
+
+        const newPrefix = prefix ? `${prefix}.${key}` : key;
+
+        let value = Ember.get(obj, key);
+
+        if (value && value.toJSON) {
+            value = value.toJSON();
+        }
+
+        if (Ember.$.isPlainObject(value)) {
+            Object.assign(ret, squashPrefix(value, newPrefix));
+        } else {
+            ret[newPrefix] = value;
+        }
+    }
+
+    return ret;
+}
+
+/**
+ * Flatten a nested object into a single level, with dotted paths for keys
+ *
+ * @param obj {!Object} - The object to flatten
+ * @returns {Object} - The flattened object
+ */
+export function squash(obj) {
+    return squashPrefix(obj);
+}
+
+/**
+ * Extracts the unique object keys from an array of objects
+ *
+ * @param dataArray {Array<Object>} - An array of objects
+ * @returns {Array<string>} - The unique keys
+ */
+export function uniqueFields(dataArray) {
+    return [
+        ...new Set(
+            [].concat(
+                ...dataArray.map(Object.keys)
+            )
+        )
+    ];
+}
+
+/**
+ * Makes a value safe for a CSV
+ *
+ * @param val {boolean|null|number|undefined|string} - The value to serialize
+ * @returns {string}
+ */
+function csvSafe(val) {
+    const value = JSON.stringify(val);
+    return value ? value.replace(/\\"/g, '""') : '';
+}
+
+/**
+ * Converts an array of objects to a standard CSV file
+ *
+ * @param dataArray {!Array<Object>} - The rows of the CSV
+ * @param fields {!Array<string>} - The fields of the CSV array. If not supplied, the the headers are calculated based on the
+ * first row of data.
+ * @param columnSeparator {string} - A custom column separator
+ * @returns {string}
+ */
+export function writeCSV(dataArray, fields, columnSeparator=',') {
+    const csv = [fields.join()];
+
+    for (const item of dataArray) {
+        const line = [];
+
+        for (const field of fields) {
+            line.push(
+                csvSafe(item[field])
+            );
+        }
+
+        csv.push(
+            line.join(columnSeparator)
+        );
+    }
+
+    return csv.join(NL);
+}
--- a/exp-models/addon/utils/csv-writer.js
+++ b/exp-models/addon/utils/csv-writer.js
@@ -3,16 +3,16 @@ import Ember from 'ember';
 /**
  * New line string
  *
- * @type {string}
+ * @type {String}
  */
 export const NL = '\r\n';
 
 /**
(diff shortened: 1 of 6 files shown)


exp-models/addon/utils/csv-writer.js
```diff
@@ -0,0 +1,107 @@
...
+    return squashPrefix(obj);
+}
+
+/**
+ * Extracts the unique object keys from an array of objects
+ *
+ * @param dataArray {Array<Object>} - An array of objects
+ * @returns {Array<string>} - The unique keys
+ */
+export function uniqueFields(dataArray) {
```
abought: As mentioned, this is our ideal solution to the problem of finding the list of headers to use.

Can you comment on performance? A project like ISP is expecting to have to handle serialization of 5-10k records at once, on the front-end; I'd be reluctant to use this by default unless we were confident it wouldn't hang the browser on an old university laptop.

exp-models/addon/utils/csv-writer.js
```diff
@@ -0,0 +1,107 @@
...
+ * Converts an array of objects to a standard CSV file
+ *
+ * @param dataArray {!Array<Object>} - The rows of the CSV
+ * @param fields {!Array<string>} - The fields of the CSV array. If not supplied, the the headers are calculated based on the
+ * first row of data.
+ * @param columnSeparator {string} - A custom column separator
+ * @returns {string}
+ */
+export function writeCSV(dataArray, fields, columnSeparator=',') {
+    const csv = [fields.join()];
```
abought: Should the headers row also respect the `columnSeparator`, for consistency with line 102?

exp-models/addon/utils/csv-writer.js
```diff
@@ -0,0 +1,107 @@
...
+
+    for (const item of dataArray) {
+        const line = [];
+
+        for (const field of fields) {
```
abought: Can L93-98 be rewritten more concisely as `const line = fields.map(field => csvSafe(item[field]))`, or similar?

exp-models/tests/unit/utils/csv-writer-test.js
```diff
@@ -0,0 +1,160 @@
...
+            {
+                alpha: three,
+                'bravo.charlie': three
+            }
+        ]
+    );
+});
+
+test('uniqueFields', function(assert) {
+    assert.deepEqual(
```
abought: If we do go ahead with using unique fields: do we need to consider nesting? Or do we simply always assume that `uniqueFields` will already be squashed?

exp-models/tests/unit/utils/csv-writer-test.js
```diff
@@ -0,0 +1,160 @@
...
+
+    const squashedData = data.map(squash);
+    const fields = uniqueFields(squashedData);
+    const result = writeCSV(squashedData, fields);
+
+    assert.strictEqual(
+        result,
+        `a,mic.check${NL}"one","two"${NL}"one","two"`
+    );
+});
```
abought: Good tests so far!

Because this is schemaless data, the presence, number, and type of fields in a given record isn't really guaranteed- which admittedly is an unusual challenge for a CSV file!

There are a few edge cases where tests might be helpful to catch situations that we could reasonably expect to happen in production. Would it be ok to ask for the following tests? (and any others you can think of along these lines)

1. Serialize 2-3 records that have some, but not all, of the same fields as each other. Make sure the gaps are handled correctly, and that the correct fields line up across rows.
2. Serialize 2-3 records with a pre-provided list of headers, including some fields that are not present in one or more of the given records.
3. Serialize 2-3 records that have different *kinds* of data in each record. For example, where a record was sometimes a leaf entry (null), but other times was an object or an array. This test is important because the `JSON.stringify` representation of the latter cases would break the CSV file, unless the whole thing was deliberately coerced to a string.  Serializing an entire variable-length array into a single cell of a spreadsheet isn't ideal, but it's about the best compromise we've got when dealing with the restrictions of a tabular format.

exp-models/addon/utils/csv-writer.js
binoculars: Yes

exp-models/addon/utils/csv-writer.js
```diff
@@ -0,0 +1,107 @@
...
+    return ret;
+}
+
+/**
+ * Flatten a nested object into a single level, with dotted paths for keys
+ *
+ * @param obj {!Object} - The object to flatten
+ * @returns {Object} - The flattened object
+ */
+export function squash(obj) {
```
binoculars: Yes, to handle map. Will add comment.

exp-models/addon/utils/csv-writer.js
```diff
@@ -0,0 +1,107 @@
...
+        }
+    }
+
+    return ret;
+}
+
+/**
+ * Flatten a nested object into a single level, with dotted paths for keys
+ *
+ * @param obj {!Object} - The object to flatten
```
binoculars: My mistake, it looks so similar to JSDoc, I thought it was. Exclamation in JSDoc mean a required parameter. Will change it to YUIDoc syntax.

exp-models/addon/utils/csv-writer.js
binoculars: Tried that. Linter error for creating a function within a loop 🙄

exp-models/addon/utils/csv-writer.js
binoculars: One could simply not use this if the fields are already well defined--which is why I separated it out into it's own function. If you just want to parse the first row,

```js
[dataArray[0]]
```

could be passed to it

abought: That is not an untrue statement- but if we want to consider using this for ISP, it's a question that we may need to examine fairly soon regardless. 

I realize that the front-end emphasis on logic such as this is a fairly unusual constraint. Alas.

binoculars: How about a maxRows optional argument as a tradeoff? If present, it would only parse the first n rows.

abought: That would put a bound on it, but do we have evidence to know whether this needs optimizations at all? If we don't have benchmarks yet, I'm open to proposals for a reasonable testcase that involves 10k records. 

It's possible that a simplified testcase would reveal other bottlenecks entirely.

exp-models/tests/unit/utils/csv-writer-test.js
binoculars: I've got 1 and 2, but can you explain how 2 is different than 4?

binoculars:
Example output file imported into Google Sheets: screen shot 2016-12-08 at 13 36 01

[test_file.csv](https://github.com/CenterForOpenScience/exp-addons/files/640281/test_file.csv) (Had to attach it as txt)

exp-models/tests/unit/utils/csv-writer-test.js
```diff
@@ -120,16 +120,44 @@ test('uniqueFields', function(assert) {
...
+            test2: one
+        },
+        {
+            test1: two,
+            test2: two
+        }
+    ];
+
+    assert.strictEqual(
+        writeCSV(data, ['test1', 'test2'], '\t'),
```
abought: Good to verify that the separator argument works!

For the next step (adding to experimenter), we're going to do away with the TSV output option altogether and replace it with CSVs; it was really just there until we had the code to write CSV files correctly.

(The only difference is the delimiter; using tab-delimited let us build a quick prototype without worrying about edge cases. CSV is more familiar to researchers though)

exp-models/addon/utils/csv-writer.js
```diff
@@ -0,0 +1,124 @@
...
+ * @return {String}
+ */
+function csvSafe(val) {
+    let value = val;
+
+    switch (typeof value) {
+        case 'boolean':
+            return value ? 'TRUE' : 'FALSE';
+        case 'number':
+            return value.toString();
```
abought: As I've said before, I really appreciate the thought you put into checking these various edge cases.

PR #211, opened 2016-12-21
Add page numbers to top of forms [LEI-331]

binoculars:
Refs: https://openscience.atlassian.net/browse/LEI-331
Companion to: CenterForOpenScience/isp#118

## Purpose
Adds page numbers to top of from

## Summary of changes
Created a new component for displaying page numbers, and added it to each of the components that require it

## Testing notes
It should display the correct page number, even when the session is resumed


exp-player/addon/components/exp-page-number/template.hbs
```diff
@@ -0,0 +1 @@
+<p style="width:100%; text-align: right">{{pageNumber}}</p>
```
abought: In general we try to avoid use of hardcoded styles in templates, but I realize this is kind of an unusual situation...

Note solely for future reference... ember components are rendered in an outer div which is slightly unnecessary here. This can be controlled in component JS, eg with the `tagName` and `classNames` properties. (normally I wouldn't bother with those, but for a simple component like this it might be useful)

exp-player/addon/components/exp-page-number/component.js
```diff
@@ -0,0 +1,7 @@
+import Ember from 'ember';
```
abought: Small request: could you rename this component?

The `exp-` prefix has a very special meaning in Experimenter: it means that something can be used as a frame of an experiment. Other utility components, like radio-group, do not use this prefix for that reason.

https://github.com/abought/exp-addons/blob/ece08446e342d9cc888123fa897fdeaed35e89ef/exp-player/addon/utils/parse-experiment.js#L79-L79

binoculars:
I think I got everything--let me know if anything is missed.

PR #212, opened 2016-12-21
Fix average label positioning

binoculars:
Refs: https://openscience.atlassian.net/browse/LEI-296
Companion to: will update ISP submodule reference after merge

## Purpose
Fixes "Average" label positioning for question 5

### Before:
screen shot 2016-12-21 at 15 03 14
screen shot 2016-12-21 at 15 03 23

### After:
screen shot 2016-12-21 at 14 58 11
screen shot 2016-12-21 at 14 58 35

#### x-pseudo
screen shot 2016-12-21 at 15 24 17
screen shot 2016-12-21 at 15 24 36

#### German
screen shot 2016-12-21 at 15 25 52
screen shot 2016-12-21 at 15 26 03

#### Indonesian
screen shot 2016-12-21 at 15 26 45
screen shot 2016-12-21 at 15 26 58

## Summary of changes
CSS only fix for the exp-overview scss file

## Testing notes
The average label on should be positioned between 5 and 6 above the radio buttons

### DIFF
--- a/exp-player/addon/styles/components/exp-overview.scss
+++ b/exp-player/addon/styles/components/exp-overview.scss
@@ -1,5 +1,6 @@
 .format-average-label {
-    margin-left: 10px;
-    position: absolute;
+    display: block;
+    width: 100%;
+    margin-left: calc(50% + 10px);
     top:20px;
 }


exp-player/addon/styles/components/exp-overview.scss
```diff
@@ -1,5 +1,6 @@
 .format-average-label {
-    margin-left: 10px;
-    position: absolute;
+    display: block;
+    width: 100%;
+    margin-left: calc(50% + 10px);
```
abought: Can you comment on the meaning and origin of these numbers?

The big challenge we've always had for ISP is internationalization. How does it look in the three translations we have currently implemented (German, indonesian, and twice-as-long-a-word "x-pseudo"?

binoculars: display block / width 100% makes it assume the width of the parent, then we need to move it over 50% + the extra margin of the radio button.

See updated main comment for other languages

abought:
The underlying CSS is weird, but I didn't notice any major new side effects and it's most definitely an improvement over the original behavior. Passing to QA so they can whack on it with a hammer.

PR #213, opened 2017-01-04
Conditionally check for hashed password in serializer

binoculars:
Refs: https://openscience.atlassian.net/browse/LEI-281
Companion to: CenterForOpenScience/experimenter#132

## Purpose
UI Responsiveness in account creation

## Summary of changes
Adds conditional check for encrypted password


binoculars:
Checks for hashed password and runs bcrypt if the password is not hashed, otherwise, it just calls super

PR #217, opened 2017-01-10
[hold][WIP] Suppress questions in sensitive experimental sites

abought:
Refs: https://openscience.atlassian.net/browse/LEI-344
Companion to: https://github.com/CenterForOpenScience/isp/pull/127

⚠️ Hold until an arabic translation is ready. This will need very thorough and careful testing to be considered for release.

## Purpose
Exclude certain questions from surveys if the locale is marked sensitive (based on ISP-specific config and ISP-specific currentUser service extension)

## Summary of changes
tbd

## Testing notes
- [ ] Make sure that conditional questions (like "which religion" on the first survey page) still show and hide correctly, even if using a site marked as sensitive.
- [ ] Make sure that marked questions are excluded when using a site ID marked as sensitive.
  - [ ] Try everything you can possibly do to this page to get the field to show when it shouldn't. Sneak in via the back button or URL hacking, break outside the user flow, come back later when already logged in- charge the gates.
- [ ] Make sure that data is serialized correctly, eg integer fields are still integers

### DIFF
--- a/exp-player/addon/components/exp-overview/component.js
+++ b/exp-player/addon/components/exp-overview/component.js
@@ -63,7 +63,7 @@ const questions = [
         scale: [
             {label: 'survey.sections.1.questions.2.options.male', value: 1},
             {label: 'survey.sections.1.questions.2.options.female', value: 2},
-            {label: 'survey.sections.1.questions.2.options.other', value: 3},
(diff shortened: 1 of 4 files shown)


abought:
Rather than filtering questions, client has elected to go a different way and use alternate translations entirely for arabic. Closing this accordingly.

PR #225, opened 2017-02-17
[wip] Add an "unsafe" save base frame for Lookit use case

abought:
Refs: https://openscience.atlassian.net/browse/LEI-369
Companion to: tbd

## Purpose
We recently pushed a fix that did not allow a user to advance through the study until the server confirmed that save had succeeded.

As it turns out, however, going fullscreen counts as a user interaction. When we used fire and forget saves this was fine... but triggering `showFullscreen` from within a promise does not count as coming from a user event handler. 

Solution: let certain, very specific lookit frames opt in to an "unsafe" save mode, which uses the old fire-and-forget style. This lets them auto-advance, and relies on a "safe" frame at the end  (like an exit survey) to make sure that data got sent to the server.

It's not perfect but at least for now provides a workaround to accommodate the workflow of both ISP and lookit.

## Summary of changes
- [x] Add new `exp-frame-base-unsafe` component
- [x] Change select lookit-specific components to use this as a base class
  - [x] Retrofit lookit frames that have manual save behavior, to work with the new system
  - [ ] Crosscheck new kim frames to be sure the conversion covers code coming soon

## Testing notes
- Confirm that ISP (with this addon) work as intended (regression test to be triggered at some future date for next isp release that requires a change to addons)
- Full check of Lookit experiment flow
  - [ ] Go through an experiment. Find a way to allow videos to load while preventing data from saving; verify errors go to console but experiment can continue
  - [ ] Generate a network failure condition during the last frame (exit survey) and confirm that user can't move forward there unless data successfully saves. (error message + next button fails to act)
  - [ ] Noticed that videos are saving very, very quickly- make sure to check the server and verify videos are well formed and saving correctly.


abought:
@kimberscott This PR may affect your new code. I've done a pass of manual regression testing, but let me know if anything else breaks. 

Your new PR adds new frames. If it would save you time and stress, I can help convert over to the new rules when things are ready.

Major rule changes:

- Added a new `Unsafe` base frame, which allows a form of saving that can lead to dataloss (but does not interfere with letting the next page automatically go full screen without user intervention)
  - It also lets you advance to the next page without error messages that would interrupt or take the user out of the workflow.
- Any frame that you intend to appear BEFORE something that goes fullscreen will  need to inherit the `Unsafe` version of a base frame. [example](https://github.com/CenterForOpenScience/exp-addons/pull/225/files#diff-3383e1867e642c2d870d630c6034ea7dR4)
  - Because of the "frame before" rule: if you expect two of a given frame to be used in a row, then the given type will also need to use the unsafe base class
  - The `Unsafe` version also avoids interrupting auto-advance in general. By default, the new `exp-base-frame` does not let someone advance to the next frame if the server failed to save their JSON data
- The primary mechanism for preventing data loss is to assume that the last frame (exit survey) is a stickler for making sure the save happened. If you disconnect your wifi, then hit "continue" on the exit survey with your responses, it should show a red message and prevent you from seeing the "thank you now then exit" page.

kimberscott:
Thanks! We should already be using this in the affected geometry frames: 60e4f355e9662e004f6477601c4ca210e0566103

PR #227, opened 2017-02-27
[rfr] Sample YUIDoc boilerplate

abought:
Refs: https://openscience.atlassian.net/browse/LEI-377

## Purpose
Add the ability to create auto-generated text documentation, especially for user-facing frames.

## Summary of changes
Add YUIDoc configuration and build commands. Add a demonstration of recommended practices for documenting three components:
- The base frame
- One frame that inherits the base, without changing input parameters or serialized data
- One frame that inherits the base, and changes both input parameters and serialized data

Also improve existing documentation to the minimum standard needed for it to show up in generated docs and build correctly. 

screen shot 2017-02-27 at 1 06 08 pm

### Rules
YUIDoc is based solely on the docstrings you give it. If you omit a docstring, a method won't appear in the docs. If you use annotate a method as a property, it will go with what you tell it.

It will often ignore method or property declarations unless you first specify a class.  Likewise, it may omit classes that do not define an associated module at the top of the file. Since most frames `@extend` a class, in practice this works fine.

YUIDoc insists that its docstrings have a very specific [syntax](http://yui.github.io/yuidoc/syntax/): 
```
/**
 * First line must have two asterisks, to differentiate documentation from 
 * other sorts of multiline source code comment
 */
```
In the strictest sense, not every basic JS concept translates perfectly into the way experimenter does things. To shoehorn this into something we can document, I have adopted a convention that all configurable frame properties are documented as `@property` strings, and all fields that are captured and serialized as part of data collection are defined as `@param`s of the `@method serializeContent`. Any frame that collects data will need to override `serializeContent`.

(note that *every* frame will capture `eventTimings`. We are definitely abusing YUIDoc syntax for this method!)

## Testing notes
Within the `exp-player` directory, you can preview the documentation (and auto-rebuild when docstrings change) by running the following command:

`yuidoc --server`

Then visit a specific page, like the `frames` section of the documentation:
`http://127.0.0.1:3000/modules/frames.html`

## Deployment notes
The simplest path to deployment would be using [github pages](https://help.github.com/articles/configuring-a-publishing-source-for-github-pages/) to deploy built documentation. We will hold off on doing this for now so as to avoid exposing preliminary/ unmaintained docs that are a work in progress.

Also considered [readthedocs](http://read-the-docs.readthedocs.io/en/latest/getting_started.html#write-your-docs), but this does not appear to support YUIDoc. It works well for python tools like sphinx, and is fine for prose, but doesn't seem to play as well with javascript documentation.

### DIFF
--- a/exp-player/addon/components/exp-frame-base-unsafe/component.js
+++ b/exp-player/addon/components/exp-frame-base-unsafe/component.js
@@ -1,12 +1,31 @@
 import ExpFrameBaseComponent from '../../components/exp-frame-base/component';
 
 /**
+ * @module exp-player
+ * @submodule frames
+ */
+
+/**
  * A modified version of the base frame, specifically designed to allow "unsafe" saves and let the user
  *   advance even if data from this frame did not save
  *
  * The primary use case for this component is experiments that need to move through a series of frames without leaving
  * fullscreen mode: if next is fired as a promise, it does not count as a user interaction event
  *  https://openscience.atlassian.net/browse/LEI-369?focusedCommentId=46133&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-46133
+ *
+  * This frame has no configuration options because all of its logic is internal, and is almost never directly used
+ *   in an experiment. It exports no data. Sample experiment definition usage (provided for completeness):
+ ```json
+ "frames": {
+       "my-sample-frame": {
+         "kind": "exp-base-frame-unsafe"
+       }
+    }
+ * ```
+ *
+ *
+ *  @class ExpFrameBaseUnsafe
+ *  @extends ExpFrameBase
  */
 export default ExpFrameBaseComponent.extend({
     actions: {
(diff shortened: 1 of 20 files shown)


abought:
Initial boilerplate set up. Passing to @kimberscott for review along with some notes and examples for how all this would work. Will hold merging until then.

kimberscott:
To run `yuidoc --server`, what do I need to install? (Is this a package that can just be installed via yarn/npm? I've run `yarn install --pure-lockfile` after pulling these changes and `yarn add yuidocjs` but still get an error `yuidoc: command not found`.)

abought:
> Update: npm -g install yuidocjs as indicated here works fine, and I see the documentation! But should I be avoiding doing that directly instead of using yarn?

Yeah, apologies... the `-g` means global, so I didn't catch this dependency. (since it was already on my system) 

I'll fix that as part of any other cleanup before this gets merged.

abought:
Also, as a general note, commit ee01d054da3b67fdfd2429132ef0f707cac742cb is deliberately isolated to show the boilerplate required for a single general-purpose frame. Hope that helps. 🐠

abought:
I know you mentioned you were adding docs to your geometry PR based on this. Would it be helpful if I went ahead and merged this scaffolding to have available? I can wait if you'd prefer.

kimberscott:
Sure, that'd be great! (I'd just pulled these changes into the geometry branch.)

PR #229, opened 2017-03-16
Final code for geometry study

kimberscott:
## Purpose

Finalize code for geometry alternation study ahead of starting testing

Ready to be merged into develop branch & then pushed to production!


abought:
Changes look small and ok.

We haven't done any sort of QA since before the geometry study went in, and merging this fast could have unknown effects on production. Is this a concern?

If so, we can send it only to staging first to be tried out live, before it hits actual users. Any preference?

After merging, I will need to send a separate change to the lookit repo to use this new code. I will wait to do that pending response on the above.

kimberscott:
I'm okay with just sending it to production; I've done very basic checking of the physics study on staging, and in principle the changes for the geometry study shouldn't affect it (with the exception of the fullscreen cursor behavior, which if anything fixes a potential existing bug in the physics study too). Thanks for your caution, though!

PR #231, opened 2017-03-17
support for checking whether participant is younger than age range

kimberscott:
Companion to: 

## Purpose
Provide function to check whether a potential participant is younger than the listed age range, to support showing different warning messages (asking people to wait if the child is younger)


exp-models/addon/models/experiment.js
```diff
@@ -139,6 +139,21 @@ const Experiment = DS.Model.extend(JamModel, {
         maxDays = maxDays || Number.MAX_SAFE_INTEGER;
         return minDays <= age && age <= maxDays;
     },
+    isOldEnough(participant) {
```
abought: I notice that a lot of this code is similar to `isEligible`, which could add challenges for maintenance later. Would you object if I sent a PR to your branch, demonstrating an idea to help simplify this?

(cool github tidbit: you can send a PR to another PR! The message will show up in your fork, and if you merge it, my commits would automatically become part of this here thread. It's very meta)

kimberscott: That would be great! Yes, this is just a pared-down isEligible, maybe it makes more sense to return in/out PLUS older/younger or -1/0/1 rather than a boolean here, or something else.

abought: Yup, we're thinking on similar lines.

- We try to avoid changing the meaning of methods that are already used because it can break things... (esp since "in range" would be 0, and we don't want a boolean method returning a falsy value for "yes")
- But a helper function would abstract away the code. -1 for below, 0 for in, 1 for above: this follows the rules for sort functions, because someday it would give an easy way to sort the list of children by eligibility without much extra work.... ;)

abought: PR sent to Kim. Will hold this one until she's had a chance to look over proposed changes, and gives the all-clear for both this and the Lookit companion PR.

kimberscott:
Should be all set, tested out locally--thanks so much, Andy!

PR #232, opened 2017-03-21
[rfr] Make timing event payloads customizable, and apply to sample video mixin

abought:
Refs: 
https://openscience.atlassian.net/browse/LEI-391
https://openscience.atlassian.net/browse/LEI-392

## Purpose
Make it possible for some frames to *always* customize the event payload with custom parameters used for every timing event captured.

Demonstrate use of this feature with the video recorder mixins

## Summary of changes
- Add `makeTimeEvent` hook to customizable base frame
- Use in mixins
- Add helper methods to VideoRecordMixin, and update most frames that currently use it
- Update docs for current best practices on these events

### TODO
- [x] Refactor videorecorder mixin responsibilities so it can capture stream time
- [x] Manual testing of events (in euler + physics experiments)
- [x] Add promise return values and throw annotations to helpers where warranted by underlying service
- [x] Update and expand final documentation for events
- [ ] Feedback round to verify event names (for compatibility with existing result coding toolchain)

## Testing notes
Light manual testing: on staging, compared the euler and baby physics events lists (before and after). Tried pausing videos, entering/exiting fullscreen, etc to confirm boilerplate was being captured correctly.

After this change, we will now capture slightly more events than before. (some old frames were less verbose, but now they are more consistent)


abought:
1. Per original request, would this sample proposal meet the needs?
> VideoRecorder mixin should overwrite the standard setTimeEvent action with one that includes the videoId and streamTime
Rather than: each frame using this mixin includes its own sendTimeEvent function instead.

2. Question for @kimberscott on second half of the request: clarifying which "general video timing events" we want to put in the generic mixin.
> Record events about video recording directly in the `VideoRecordMixin`, rather than handling those from any frames using it.

Each frame has some latitude in what it records or how; currently there is no 100% generic "video recording events" use case. I propose to add the very most common things to the VideoRecordMixin, which will add timing events boilerplate too.

Proposed "video recorder" helpers (let me know if we want other boilerplate)
- `setRecorder` (just make sure `recorder` can be found in a consistent place)
- `startRecording(recorder)`
- `pauseRecording(recorder)`
- `stopRecording(recorder)`

Other recorder events (like hasCamAccess) get set as part of creating the recorder, and not every frame does that the same way.

kimberscott:
Looks good! That set of events for video recorder events makes sense to me. 

I see the issue about the onCamAccess, onConnectionStatus, etc. events currently getting set up only upon creating the recorder, but although the recorder may be created differently, the actual events to record are pretty standard. Is it possible for event-recording to be part of the standard handlers for those events? It looks like in exp-player/addon/services/video-recorder.js, for instance, recording an appropriate event could happen in _onConnectionStatus, _onCamAccess, etc.

abought:
Thanks for the feedback. Not every frame was consistent previously, but I can make these part of the standard package.

There's a fairly good chance that the required refactorings will collide with your latest PR, and it will require some careful vetting to check for subtle recording bugs. 

Though I plan to finish this today or tomorrow, how would you like to plan the merging timeline? I'll tentatively plan to hold this PR until you give the all clear, so that we are sure that we have time to start experiments without untested refactorings at the 11th hour.

kimberscott:
Yep, this will definitely require some changes to existing frames (at least to avoid redundant event records) but it's worth it going forward. Shall we plan to merge these changes next week? Once you've set up for these events to be recorded as part of the videoRecorder mixin, I can go back through to all of the affected frames (off the top of my head, video-preview, video-physics, geometry-alternation, and the three new frames for the round 2 studies) to make the necessary changes, and double-check that we don't break anything on the currently-running studies (physics & geometry).

abought:
Changes updated to give a sense of the new mixin methods and will hold this until there are no conflicts for @kimberscott 's work. I avoided preemptively altering the list of events for compatibility with Kim's existing analysis workflow, but am happy to make changes as needed.

If you're using experimenter, you can compare the "before" picture (`bb0sT, 3/21/2017`)  to the "after" payload (profile `eOiv3, 3/23/2017`), from this link.
https://staging-experimenter.osf.io/experiments/58d015243de08a00400316e0/results?sort=-modified_on

abought:
I've set a note to revisit this PR on Thursday: do you think this is something that could be merged and vetted this week? 

I know schedule will be tight, but if we can keep PRs from going stale (without breaking things for collaborators), that's always helpful.

abought:
Also, I'll work on resolving the merge conflicts.

kimberscott:
Yes! Sorry we had to prioritize merging in the other changes, but I do have scheduled to look through and adjust the existing frames this affects on Wednesday.

abought:
> Yes! Sorry we had to prioritize merging in the other changes, but I do have scheduled to look through and adjust the existing frames this affects on Wednesday.

You made the right call by waiting; no worries. 

If you'd like, I've got some time free before 2pm, and could work on converting the new frames along similar lines as what's already in this PR. Just let me know.

kimberscott:
If you're up for it, that'd be great - thanks so much! (I think they should mostly have TODO comments already where they're recording events that got moved.)

abought:
> If you're up for it, that'd be great - thanks so much! (I think they should mostly have TODO comments already where they're recording events that got moved.)

@saradbowman could answer better than I, but I believe the plan is increased availability this week due to the sensitive timeline. 

Until I get any set list of specific tasks from her, I'll proceed on the assumption that priority is to help you with your general queue to get collaborators on board by Friday. :)

abought:
Updated to work with newest frames. Passing to Kim for any review she'd like to do- in particular, to make sure this preserves the intended behavior of the frames. Will hold until she's had time to look at this.

(I did a light pass but am running light on context for requirements)

Sorry for the documentation change spam in this diff. Since filenames are just alphabetical, just scroll down until you get to something that doesn't have `docs/` in the path and start reading from there. :)

kimberscott:
Finally working on review now! :)

kimberscott:
I've tried this out and made some minor changes to the individual frames. I tried making a PR to your abought:feature/lei391-392-event-recording branch so you could add in my changes, but let me know if there's a better way to do that! 

The only outstanding issue for us is that the recorderReady events that each frame sends like this

```
const installPromise = this.setupRecorder(this.$('#videoRecorder'), true, {
    hidden: true
});
installPromise.then(() => {
    this.send('setTimeEvent', 'recorderReady');
});
```

are never recorded in any frames, which makes me suspect I don't understand how the installPromise works. Any ideas about what might be going on there? From our current data, it looks like this event hasn't been recorded before, so it's not a new thing, but it might be good to address while we're here.

abought:
Good catch on the recorderReady event.

It looks what is happening is that the `resolve` method specific to the installPromise is never actually getting resolved, if you choose to record:
https://github.com/abought/exp-addons/blob/0dece495326bc05fbc36ad5041d123717a718a44/exp-player/addon/services/video-recorder.js#L164-L164

Instead if returns a different promise. Normally this all works by magic, because under the hood some other code has done the work of calling resolve for us. But in this case not so much. Theoretically the fix will be pretty simple but developing a testing plan would be another matter; I've not been able to find any tests or spec for this code. 

I suggest merging what we have here and then evaluating that as a separate PR, per Sara decision? (I'd be more sanguine about changing that piece if our schedule weren't so tight, but trying to play it safe for collaborators)

kimberscott:
Thanks for looking into that, and cool that there's a simple explanation!

Sounds fine to me to merge this and deal with the installPromise separately - all we do with the promise is record the event anyway, and this code hasn't changed anything about that not working :) It'd probably be cleaner, going forward, to actually get back a promise, and be able to use that (rather than videoStreamConnection + hasCamAccess) to decide when to start trials.

Up to Sara on timeline!

exp-player/addon/components/exp-lookit-geometry-alternation/component.js
```diff
@@ -550,7 +550,7 @@ export default ExpFrameBaseUnsafeComponent.extend(FullScreen, VideoRecord,  {
          *
          * @event startIntro
          */
-        _this.sendTimeEvent('startIntro');
+        _this.send('setTimeEvent', 'exp-alternation:startIntro');
```
abought: @kimberscott : With the new `makeTimeEvent`, this means that the event name will be:

`exp-alternation:exp-alternation:startIntro`

Is that intentional? I left some usages in old frames like this, but refactoring hinges on whether there is already live data  / an analysis pipeline for studies using this frame. New frames may give us more leeway.

Willing to make the change, but I run any changes to live experiments by you first.

kimberscott: Whoops, that is not intentional - if you're up for changing this that'd be great. None of the analysis pipeline is set up yet anyway, so no problem making it robust to a few different names for the event.

abought:
Merged changes, and they should now be on staging. I'll be switching back to other tasks now, but if there are remaining items to be addressed as part of this ticket, please let Sara or myself know.

kimberscott:
Sounds good - thanks so much!

PR #235, opened 2017-03-31
Supporting round 2 development

kimberscott:
Two small changes based on collaborators' needs - 

- allow exp-lookit-story-page (and exp-lookit-dialogue-page since it uses the same setup) to control formatting of parent text, in order to show more text if needed (for Lisa)
- create a randomizer to randomly permute a set of frames (allowing stimuli to be presented in random order) (for Bria)

Tested locally to make sure the exp-lookit-story-page/dialogue-page changes don't introduce errors with existing definitions, and to make sure randomizer `permute` works and can be nested within `random-parameter-set`.

### DIFF
--- a/exp-player/addon/components/exp-lookit-dialogue-page/component.js
+++ b/exp-player/addon/components/exp-lookit-dialogue-page/component.js
@@ -620,7 +620,7 @@ export default ExpFrameBaseUnsafeComponent.extend(FullScreen, VideoRecord,  {
 
     parentTextStyle: Ember.computed('parentTextBlock', function() {
         var parentTextBlock = this.get('parentTextBlock') || {};
-        var css = parentTextBlock['css'] || {};
+        var css = parentTextBlock.css || {};
         var cssString = '';
         $.each(css, function(propName, value) {
             cssString += new Ember.Handlebars.SafeString(`${propName}:  ${value};`);
--- a/exp-player/addon/components/exp-lookit-dialogue-page/component.js
+++ b/exp-player/addon/components/exp-lookit-dialogue-page/component.js
@@ -618,16 +618,6 @@ export default ExpFrameBaseUnsafeComponent.extend(FullScreen, VideoRecord,  {
 
     },
 
-    parentTextStyle: Ember.computed('parentTextBlock', function() {
-        var parentTextBlock = this.get('parentTextBlock') || {};
-        var css = parentTextBlock.css || {};
-        var cssString = '';
-        $.each(css, function(propName, value) {
-            cssString += new Ember.Handlebars.SafeString(`${propName}:  ${value};`);
-        });
-        return cssString;
-    }),
-
     // Utility to expand stubs into either full URLs (for images) or
     // array of {src: 'url', type: 'MIMEtype'} objects (for audio).
     expandAsset(asset, type) {
@@ -685,6 +675,10 @@ export default ExpFrameBaseUnsafeComponent.extend(FullScreen, VideoRecord,  {
 
         this.set('images', images);
         this.set('audioSources', audioSources);
+
+        var parentTextBlock = this.get('parentTextBlock') || {};
+        var css = parentTextBlock.css || {};
+        $('#parenttext').css(css);
 
         this.send('showFullscreen');
         $('#nextbutton').prop('disabled', true);
--- a/exp-player/addon/components/exp-lookit-dialogue-page/component.js
+++ b/exp-player/addon/components/exp-lookit-dialogue-page/component.js
@@ -58,7 +58,11 @@ let {
             "parentTextBlock": {
                 "title": "Parents!",
                 "text": "some instructions",
-                "emph": true
+                "emph": true,
+                "css": {
+                    "color": "red",
+                    "font-size": "12px"
+                }
             },
             "images": [
                 {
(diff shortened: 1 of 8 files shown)


abought:
I can merge this today when it's ready if needed, but bear in mind it won't get much in the form of QA from our end due to the present time constraints. (we always try to warn you when that will be the case)

 Were you and Sara able to work out a procedure for bug reporting while you're out?

exp-player/addon/randomizers/permute.js
```diff
@@ -0,0 +1,123 @@
...
+        frames.push(...thisFrame);
+    }
+
+    /**
+     * Parameters captured and sent to the server
+     *
+     * @method conditions
+     * @param {Object[]} frameList the list of frames used, in the final shuffled order
+     */
+    return [frames, {'frameList': array}];
```
abought, line 121: Conceptually, this seems similar to the goal of the `random` randomizer: take a list of predefined frame names, and return them in a random order.

Would it be correct to say that the difference is that this doesn't use a static set of choices, but rather dynamically generates the options before shuffling?

kimberscott: I think `random` actually just chooses one random frame, and this is more similar to `shuffle`. The only difference is that we apply `commonFrameProperties` to all of the frames as a convenience (figuring that often the only thing varying across the frames will be e.g. what image to show). 

I was hesitant to simply `shuffle` in case anyone was using it as-is, but it looked like it was using a different set of arguments to randomizer. If no one's using it, it's fine by me to rename this 'shuffle'!

exp-player/addon/components/exp-lookit-story-page/component.js
```diff
@@ -336,6 +347,16 @@ export default ExpFrameBaseUnsafeComponent.extend(FullScreen, VideoRecord,  {
         }
     },
 
+    parentTextStyle: Ember.computed('parentTextBlock', function() {
+        var parentTextBlock = this.get('parentTextBlock') || {};
+        var css = parentTextBlock.css || {};
+        var cssString = '';
+        $.each(css, function(propName, value) {
+            cssString += new Ember.Handlebars.SafeString(`${propName}:  ${value};`);
```
abought: I'm a bit nervous about this line; we generally try to avoid using SafeString.

The reason is that it totally trusts the content to be rendered exactly as HTML- you need to be sure that you 100% trust every person with access to any field that is part of the safestring, both now, and for any future possible use or adaptation of this component/code/template. Otherwise, they can insert arbitrary code of their own into every page you view (JS that sends API calls with your credentials, ad trackers, etc). 

The formal term is [XSS](https://www.owasp.org/index.php/Cross-site_Scripting_(XSS)), and this risk is the reason why most web frameworks escape all text by default. Is there a way to avoid this? For example, would [jquery css method](http://api.jquery.com/css/) meet your needs?

exp-player/addon/components/exp-lookit-dialogue-page/component.js
```diff
@@ -607,6 +618,16 @@ export default ExpFrameBaseUnsafeComponent.extend(FullScreen, VideoRecord,  {
 
     },
 
+    parentTextStyle: Ember.computed('parentTextBlock', function() {
+        var parentTextBlock = this.get('parentTextBlock') || {};
+        var css = parentTextBlock.css || {};
+        var cssString = '';
+        $.each(css, function(propName, value) {
+            cssString += new Ember.Handlebars.SafeString(`${propName}:  ${value};`);
```
abought: See note elsewhere about `SafeString`. (which should more accurately be named `GeeIHopeThisIsASafeStringBecauseIAmChoosingToTrustIt`)

abought (changes requested):
Under the circumstances of what you *could* be doing right now, I am very hesitant to request changes.... but two requests /  suggestions before this is merged.

exp-player/addon/components/exp-lookit-story-page/component.js
kimberscott: Whoops, I was taken in by the name. jquery css would work fine! Would it be okay to do something like `$('#parenttext').css(propName, value);` in place of building the `cssString`?

abought: I think that would be ok, or at least a big improvement. That said: the more you let the user write code directly in the page, the more risk there is of something going weird in general. Long term you may want to consider developing user stories for a very limited set of style options. This would mitigate the security risks, and ensure the site had a less buggy, more consistent look and feel.

Note that jquery loves to give a method call multiple forms; it will also take your object directly. Might be useful.
> css( properties )
properties
Type: PlainObject
An object of property-value pairs to set.

kimberscott: Yes, that's a better plan longer-term. It might really be as simple as giving control over font size, but right before heading out for a while, I figured it might be better to cover all possible bases. It's on my list now!

kimberscott: And thanks for the tip about providing `properties`. Should be set up for these two frames now.

exp-player/addon/randomizers/permute.js
abought: Thanks for the changes; I'll go ahead and merge once CI is passing.

(on the premise that responding to feedback will only get harder as the next week goes on)

Pretty sure the old randomizers have different arguments because of lack of maintenance, along with some aspects of nesting behavior. (there are some skipped unit tests to that effect if someone ever gets curious) I don't think the other platform project uses those randomizers at all.

kimberscott: Thanks! And okay--sounds like we could eventually prune the old randomizers to keep the codebase easier to understand, especially as we add replacements for basic ones like permuting / randomly selecting / etc.

abought (approved):

PR #241, opened 2017-10-04
[SVCS-505]Color Cat9 text green on completion

AddisonSchiller:
Refs: https://openscience.atlassian.net/browse/SVCS-505

## Purpose
Change color of Cat9 text once a bucket has been completed

## Summary of changes
Styled template to do so

### DIFF
--- a/exp-player/addon/components/exp-card-sort/template.hbs
+++ b/exp-player/addon/components/exp-card-sort/template.hbs
@@ -77,14 +77,14 @@
                     <div class="row">
                         {{#each group.categories as |category|}}
                             <div class="col-sm-4 small-bucket
-                                {{unless (eq category.cards.length category.max) 'text-danger validation-error'}}
+                                {{if (eq category.cards.length category.max) 'text-success' 'text-danger validation-error'}}
                                 ">
-                                <h5 class="bucket-label visible-xs {{if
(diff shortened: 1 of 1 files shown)


exp-player/addon/components/exp-card-sort/template.hbs
```diff
@@ -78,13 +78,14 @@
                         {{#each group.categories as |category|}}
                             <div class="col-sm-4 small-bucket
                                 {{unless (eq category.cards.length category.max) 'text-danger validation-error'}}
-                                ">
-                                <h5 class="bucket-label visible-xs {{if
+                                {{if (eq category.cards.length category.max) 'text-success'}}
```
felliott: Minor: the `if` condition is the same as the `unless` condition.  This could be combined into a single if statement with a true value and a false value.

exp-player/addon/components/exp-card-sort/template.hbs
```diff
@@ -78,13 +78,14 @@
...
+                                " >
```
felliott: Super-mega-nitpick: the added space between the `"` and the `>` makes the diff bigger.

felliott (changes requested):
Just minor fixes needed.

